Issue #11, opened 2016-07-20
Add null coalescing operator (??) in template expressions


mastodon0:
Dart needs a way to handle null values in template expression like the `||` operator in JS.
The elvis operator (`?.`) doesn't help in situations like `selectedValue != null ? selectedValue : 'no selection'`. 
Since dart got the support for the `??` operator in 1.12, I think it should be a part of template expressions.

(copy of issue https://github.com/angular/angular/issues/6367 in the old project)

matanlurey:
Sounds like a decent idea. We can add it to our triage.

I imagine this isn't a huge priority only because there is a (meh) workaround in the form of the ternary operator and == null check. Definitely worth doing though!

mastodon0:
I tried to implement it myself, you can see it here https://github.com/mastodon0/angular2/tree/operator_if_null
Not quite sure about the naming though.

matanlurey:
@mastodon0: Please send us a PR. We're in the process of updating our `CONTRIBUTING.md` to explain our new sync procedure, but we'll be happy to take your feature!

matanlurey:
Nice, we got this. Thanks @mastodon0!

PR #36, opened 2016-08-03
Added If Null operator (??)

mastodon0:
If null/coalesce operator for issue #11

### DIFF
--- a/lib/src/compiler/expression_parser/ast.dart
+++ b/lib/src/compiler/expression_parser/ast.dart
@@ -74,6 +74,19 @@ class Conditional extends AST {
   }
 }
 
+/// Represents the ?? expression in Dart
+class IfNull extends AST {
+  /// Condition for the null check and result if it is not null.
+  final AST condition;
+  /// Result if the `condition` operand is null.
+  final AST nullExp;
+
+  IfNull(this.condition, this.nullExp);
+  dynamic visit(AstVisitor visitor, [dynamic context = null]) {
+    return visitor.visitIfNull(this, context);
+  }
+}
+
 class PropertyRead extends AST {
   AST receiver;
   String name;
@@ -271,6 +284,7 @@ abstract class AstVisitor {
   dynamic visitChain(Chain ast, dynamic context);
   dynamic visitConditional(Conditional ast, dynamic context);
   dynamic visitFunctionCall(FunctionCall ast, dynamic context);
+  dynamic visitIfNull(IfNull ast, dynamic context);
   dynamic visitImplicitReceiver(ImplicitReceiver ast, dynamic context);
   dynamic visitInterpolation(Interpolation ast, dynamic context);
   dynamic visitKeyedRead(KeyedRead ast, dynamic context);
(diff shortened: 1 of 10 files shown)


lib/src/compiler/expression_parser/ast.dart
```diff
@@ -74,6 +74,17 @@ class Conditional extends AST {
   }
 }
 
+class IfNull extends AST {
+  AST condition;
+  AST nullExp;
+  IfNull(this.condition, this.nullExp) : super() {
+    /* super call moved to initializer */;
```
matanlurey: This was ugly code for TS, you can remove it, just

```dart
IfNull(this.condition, this.nullExp);
```

lib/src/compiler/expression_parser/ast.dart
```diff
@@ -74,6 +74,17 @@ class Conditional extends AST {
   }
 }
 
+class IfNull extends AST {
+  AST condition;
```
matanlurey: Let's properly make both of these fields final.

lib/src/compiler/output/output_ast.dart
```diff
@@ -361,6 +365,18 @@ class ConditionalExpr extends Expression {
   }
 }
 
+class IfNullExpr extends Expression {
```
matanlurey: Ditto above, just a little documentation.

lib/src/compiler/output/output_ast.dart
```diff
@@ -361,6 +365,18 @@ class ConditionalExpr extends Expression {
...
+  Expression condition;
```
matanlurey: final both of these fields

lib/src/compiler/output/output_ast.dart
```diff
@@ -361,6 +365,18 @@ class ConditionalExpr extends Expression {
...
+  Expression nullCase;
+  IfNullExpr(this.condition, Expression nullCase, [OutputType type = null])
+      : super(isPresent(type) ? type : nullCase.type) {
+    /* super call moved to initializer */;
```
matanlurey: We probably don't need this either.

lib/src/compiler/expression_parser/ast.dart
```diff
@@ -74,6 +74,17 @@ class Conditional extends AST {
   }
 }
 
+class IfNull extends AST {
```
matanlurey: Mind leaving a tiny bit of dartdoc here:

/// Represents the `??` expression in Dart

matanlurey:
Can you double check this works with the transformer? 

(Unfortunately our tests aren't very good right now)

mastodon0:
Thank you for the advice, I hope this is more like you would have done it.
I checked the JS output, it looked like `t1 == null ? "unknown value" : t1`.

kevmoo:
Please rebase on the latest `master` – we now have travis enabled.

lib/src/compiler/output/abstract_emitter.dart
```diff
@@ -321,6 +321,16 @@ abstract class AbstractEmitterVisitor
     return null;
   }
 
+  dynamic visitIfNullExpr(o.IfNullExpr ast, dynamic context) {
+    EmitterVisitorContext ctx = context;
+    ctx.print('''(''');
```
chalin: Using triple quotes here serves no purpose. Single quotes will do, and can be used here (Line 326), as well as lines 328 and 330.

kevmoo:
@mastodon0 care to do another rebase? This should fix the tests

PR #42, opened 2016-08-08
API docs: fix markdown and dev guide links

chalin:
Main changes are to metadata directives.

### DIFF
--- a/lib/src/core/metadata/directives.dart
+++ b/lib/src/core/metadata/directives.dart
@@ -20,12 +20,15 @@ import "package:angular2/src/facade/lang.dart" show isPresent;
 /// ## Understanding How Injection Works
 ///
 /// There are three stages of injection resolution.
+///
 /// * Pre-existing Injectors:
-///   - The terminal [Injector] cannot resolve dependencies. It either throws
-///     an error or, if the dependency was specified as [@Optional],
-///     returns null.
-///   - The platform injector resolves browser singleton resources, such as:
-///     cookies, title, location, and others.
+///
+///     - The terminal [Injector] cannot resolve dependencies. It either throws
+///       an error or, if the dependency was specified as [Optional],
+///       returns null.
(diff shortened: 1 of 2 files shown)


chalin:
cc @kwalrath

kevmoo:
@kwalrath are you handling the dance into the internal source base for this?

kwalrath:
Nope, this is a @matanlurey or @ferhatb sort of thing, as far as I know. But I should make sure of that.

lib/src/core/metadata/directives.dart
```diff
@@ -722,7 +727,7 @@ class DirectiveMetadata extends InjectableMetadata {
 ///
 /// ## Lifecycle hooks
 ///
-/// When the component class implements some [../../guide/lifecycle-hooks.html]
+/// When the component class implements some [lifecycle-hooks](docs/guide/lifecycle-hooks.html)
```
kevmoo: How do these link? To which files?

chalin: They link to [angular.io/dart](https://angular.io/docs/dart). URL adjustment is done when the API docs are generated.

kwalrath: The docs also appear on dartdocs.org, albeit somewhat broken.

What can we do to avoid bad links on dartdocs.org? Can we disable the Angular doc generation there, preferably with a page that says where the docs really live?

cc @astashov

astashov: We surely do. Can you make a proper page for that, with copy and stuff? I can put it to dartdocs.org.

kwalrath: Happy to. Thanks for the quick response!

kwalrath: OK, I've suggested wording at https://github.com/astashov/dartdocs.org/issues/30. Do you want me to provide the actual HTML?

chalin:
Rebased.

PR #51, opened 2016-08-21
fix(transform): report invalid `entry_points`

### LINKED ISSUE #38: Angular Dart transformer silently ignores `entry_points` files that do not exist
Consider the following excerpt from a `pubspec.yaml`:
```yaml
...
transformers:
- angular2:
    platform_directives: ...
    platform_pipes: ...
    entry_points: SOME/FILE_DOES_NOT_EXIST.dart
```

**Current behavior**
Running `pub serve` reports no error (but then fails to identify target types for static code generation), and so the app usually fails later down the road.

**Expected/desired behavior**
It would be nice if the transformer would at least report an error or warning if the `entry_points` do not exist.

* **Angular version:** 2.0.0-beta.17
* **Browser:** Chrome 50 | Chromium 45 
* **Language:** Dart

Migrated from https://github.com/angular/angular/issues/8897.

zoechi:
print an error message to `stderr` for `entry_point` values  that don't resolve to at least one file.

fixes #38

### DIFF
--- a/lib/src/transform/common/options_reader.dart
+++ b/lib/src/transform/common/options_reader.dart
@@ -40,7 +40,7 @@ TransformerOptions parseBarbackSettings(BarbackSettings settings) {
       mirrorMode = MirrorMode.none;
       break;
   }
-  return new TransformerOptions(entryPoints,
+  var transformerOptions = new TransformerOptions(entryPoints,
       modeName: settings.mode.name,
       mirrorMode: mirrorMode,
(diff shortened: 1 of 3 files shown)


matanlurey:
LGTM, I'll get this merged tomorrow morning ⏰

kevmoo:
@matanlurey you still tracking this?

matanlurey:
Yep, @ferhatb is merging it.

Issue #55, opened 2016-08-31
Unused variable in test function after TimerWrapper facade removal


b-strauss:
I just stumbled upon this after browsing recent commits.

This function:

https://github.com/dart-lang/angular2/blob/af290b020bc4ed975f73547823315f2e8e3cf10a/test/common/forms/directives_test.dart#L45

seems to not use the timeout parameter properly. From what I can tell, it needs to be passed to the Duration.

https://github.com/dart-lang/angular2/commit/af290b020bc4ed975f73547823315f2e8e3cf10a#diff-b00a2e16fd7528fd8939e9cf6ac0debf

kevmoo:
Good find. Want to try and create a pull request to fix this?

CC @ferhatb @matanlurey

b-strauss:
Yes, I'd like to do that.

matanlurey:
Closed by #57.

PR #57, opened 2016-09-02
Fix unused timeout parameter

b-strauss:
Fixes: #55


b-strauss:
@kevmoo Do you know why this is failing? The change seems trivial.

kevmoo:
@b-strauss content shell is often borked. This looks okay to me.

kevmoo:
The timeout is actually never used. This would be a better refactoring:

```dart
asyncValidator(expected) {
  return (AbstractControl c) async {
    var res = c.value != expected ? {"async": true} : null;
    return res;
  };
}
```

PR #100, opened 2016-09-28
Updated analyzer dep min version

frankpepermans:
As of beta 22, usage of trailing commas in the source code causes the angular transformers to throw when analyzing, upgrading to 0.28.x fixes the issue.

see: https://github.com/dart-lang/angular2/issues/98

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,7 +7,7 @@ documentation: 'https://angular.io/docs/dart/latest/api/'
 environment:
   sdk: '>=1.19.0 <2.0.0'
 dependencies:
-  analyzer: '>=0.27.0 <0.29.0'
+  analyzer: '>=0.28.0 <0.29.0'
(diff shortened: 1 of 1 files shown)


kevmoo:
LGTM – @ferhatb @matanlurey ??

The risk: there are still a lot of packages that don't support 0.28.x yet – so we risk throwing folks into version lock

matanlurey:
So I'm OK with this. I think we need to take a P1 to get any package not on 0.28x up-to date (this includes dartdoc, which breaks on any project with trailing commas).

kevmoo:
Just looked. Everything still works with analyzer `0.27.6`.

I think we should hold off bumping the analyzer version until we have to.

This corner case is tough, but it's not worth breaking folks who may be using other packages – there is a LONG TAIL of folks using things that don't work with the latest analyzer.

PR #101, opened 2016-09-28
Convert examples from Typescript to Dart

kulshekhar:


kulshekhar:
Rebased & resolved conflicts

PR #110, opened 2016-09-29
Enable a test that works with js compile

kevmoo:


kevmoo:
Rolled into https://github.com/dart-lang/angular2/pull/111

PR #111, opened 2016-09-29
Skip tests that timeout on Travis with Firefox

kevmoo:
See https://github.com/dart-lang/test/issues/470

Also let folks know that tool/travis.sh requires TEST_PLATFORM set


kevmoo:
nevermind

PR #123, opened 2016-10-07
NgIf API doc update

chalin:
Minimal update to the prose, fixing the examples and adding links back into the main docs.

cc @matanlurey @vikerman @kwalrath @thso

### DIFF
--- a/lib/src/common/directives/ng_if.dart
+++ b/lib/src/common/directives/ng_if.dart
@@ -1,20 +1,19 @@
 import 'package:angular2/core.dart'
     show Directive, ViewContainerRef, TemplateRef;
 
-/// Causes the element and its contents to appear in the DOM conditionally.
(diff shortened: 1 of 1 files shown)


lib/src/common/directives/ng_if.dart
```diff
@@ -1,20 +1,19 @@
 import 'package:angular2/core.dart'
     show Directive, ViewContainerRef, TemplateRef;
 
-/// Causes the element and its contents to appear in the DOM conditionally.
+/// Causes a element and its contents to be added/removed from the DOM
```
kwalrath: a -> an

(maybe we can fix this inside elgoog?)

kwalrath: Also, Patrice showed me a rendered version of this page, which looks good.

chalin:
This change has been included, along with other API entries, in #124.

PR #136, opened 2016-10-18
Cleaned up grammar

jonahwilliams:


lib/src/modules/compiler/template_parser/GRAMMAR.md
```diff
@@ -0,0 +1,38 @@
+#Template Grammar
+
+###Todos
+* Can you interpolate comments?
```
matanlurey: Good point. I imagine it could be useful for debugging in a QA environment. I don't think it is supported yet.

lib/src/modules/compiler/template_parser/GRAMMAR.md
```diff
@@ -0,0 +1,38 @@
...
+* Can you have multiple pipes?
```
matanlurey: Meaning can you `A | pipe | pipe`? I actually don't know.

I sent an email to Victor on NgJs - let's see what they say.

lib/src/modules/compiler/template_parser/GRAMMAR.md
```diff
@@ -0,0 +1,38 @@
...
+* What is the best way to handle the additional constraints on expressions?
```
matanlurey: Good question: I don't know how much of the Dart language AST we are going to need to support "full" expressions. Maybe it's worth getting a copy of the Dart expression grammar to see what's worth supporting.

lib/src/modules/compiler/template_parser/GRAMMAR.md
```diff
@@ -0,0 +1,38 @@
...
+* What are we forgetting?
+
+
+```dart
+Node           = VoidTag | OpenTag {Node} CloseTag | Comment | Text;
+OpenTag        = "<" TagName WhiteSpace {WhiteSpace} {Attribute WhiteSpace {WhiteSpace}} ">";
+VoidTag        = "<" TagName WhiteSpace {WhiteSpace} {Attribute WhiteSpace {WhiteSpace}} "/>";
```
matanlurey: Just a nit/question/clarification: Are we saying all void tags _must_ match `<tag />'? What about implicit voids, i.e. `<br>`?

lib/src/modules/compiler/template_parser/GRAMMAR.md
```diff
@@ -0,0 +1,38 @@
...
+CloseTag       = "</" TagName {WhiteSpace} ">";
+Comment        = "<!--" {RawText} "-->";
```
matanlurey: Hmm if we structure like this it should be possible to do interpolations :)

lib/src/modules/compiler/template_parser/GRAMMAR.md
matanlurey: Just replied - the answer is yes. This is one area where it _won't_ match the Dart expression AST, so we'll need the expression syntax to be something like `Expression (| Pipe)?`.

lib/src/modules/compiler/template_parser/GRAMMAR.md
jonahwilliams: Good point, fixing this.

lib/src/modules/compiler/template_parser/GRAMMAR.md
jonahwilliams: Expression and zero or more pipes, added

lib/src/modules/compiler/template_parser/GRAMMAR.md
jonahwilliams: might want to just try to stay at parity with the current parser for now

lib/src/modules/compiler/template_parser/GRAMMAR.md
jonahwilliams: I'm not sure if its possible to modify comments in the dom though...brb

PR #148, opened 2016-10-20
Convert the lexer back to synchronous, add a simple parser

matanlurey:
The asynchronous interface was unnecessary complexity at this point.

Added a simple parser that deals with elements and text nodes for now.


matanlurey:
Ignoring content shell breakages for now.

PR #166, opened 2016-11-02
Added scanning/parsing of structural directives, basic error handler

jonahwilliams:


lib/src/modules/compiler/template_parser/test/parser_test.dart
```diff
@@ -130,5 +130,34 @@ void main() {
...
+        parse('<div *ngIf="baz" *ngFor="let foo of bars"></div>'),
+        [
+          new NgElement.unknown('template', childNodes: [
+            new NgProperty('ngIf', 'baz'),
+            new NgElement.unknown('div')
+          ])
+        ]);
+    });
+
+    test('should throw when provided an error handler on multiple structural'
```
matanlurey: Can we add a test that shows the emitted error has source span information that helps the user fix this problem?

jonahwilliams: I would consider all of the errors currently to be placeholders, I will work on some custom error classes next.

PR #167, opened 2016-11-04
Add interpolation support for lexing and parsing

matanlurey:
Closes https://github.com/dart-lang/angular2/issues/146


lib/src/modules/compiler/template_parser/lib/src/lexer/sync_lexer.dart
```diff
@@ -208,12 +208,52 @@ class _SyncNgTemplateLexer extends NgTemplateLexerBase {
     _scanText();
   }
 
+  // Scans between {{ and }}.
+  void _scanInterpolate() {
+    var char = peek();
+    while (true) {
```
jonahwilliams: What happens to the span scanner if it runs out of input? i.e. someone forgets to close a brace.  This would probably be as much context as we'd need for a `UnclosedInterpolate` error.

matanlurey: Done.

PR #220, opened 2016-11-30
Sync 2.2.0

matanlurey:
### API changes
  * Breaking changes
    * Using `@ViewQuery|Children|Content|` in a constructor is no longer valid.
      This caused significant extra code to need to be generated for a case that
      is relatively rare. Code can safely be moved into a setter in most cases.

**BEFORE**
```dart
class MyComponent {
  QueryList<ChildComponent> _childComponents;

  MyComponent(@ContentChildren(ChildComponent) this._childComponents);
}
```

**AFTER**
```dart
class MyComponent {
  QueryList<ChildComponent> _childComponents;

  @ContentChildren(ChildComponent)
  set childComponents(QueryList<ChildComponent> childComponents) {
    _childComponents = childComponents;
  }
}
```

### Bug fixes
  * Importing `angular2/reflection.dart` now works properly.


matanlurey:
One `content-shell` flake. Pushing.

PR #226, opened 2016-12-06
Sync 2.3.0 to github

matanlurey:
/cc @kevmoo


matanlurey:
@kevmoo We don't have support for pkg/build externally yet. I don't think observable is a BC?

matanlurey:
Sorry for the delay. I'm going to close this and re-sync with more changes instead.

PR #251, opened 2017-01-18
Sync to AngularDart v3.0.0-alpha

matanlurey:
I still intend to do a follow-up sync with `CHANGELOG.md` updates before releasing.


pubspec.yaml
```diff
@@ -26,10 +26,10 @@ dev_dependencies:
   args: ^0.13.6
   build_runner: ^0.1.1
   build_test: ^0.2.0
```
kevmoo: Missing the latest version here - `0.3.1`. It's syncing internally.

matanlurey: Will fix upstream.

matanlurey:
We plan to have the deprecated differs fixed as part of `3.0.0` final, but it's a big enough change where we wanted to call it out early @kevmoo.

PR #254, opened 2017-01-19
API docs (DomSanitizationService): quote inlined HTML

chalin:
Quote inlined HTML anchor element (otherwise, all the rest of the API doc entry gets marked as a link). Other copyedits.

cc @kwalrath @kevmoo


kevmoo:
@matanlurey @ferhatb wanna run w/ the sync here?

PR #256, opened 2017-01-25
Use forEach less, added some finals

kevmoo:


adaojunior:
Hi, I'm curious, why are you switching forEach to for ?

kevmoo:
> Hi, I'm curious, why are you switching forEach to for ?

`for...each` is easier to optimize. There is no closure. The compiler doesn't have to worry about `.forEach` being overridden in a subclass.

kwalrath:
Should we mention this in the docs?

matanlurey:
Maybe for the contributing guide, but this isn't a requirement users.

Issue #260, opened 2017-01-31
Async pipe with stream does not seem to work


daniel-v:
Example app: https://gist.github.com/daniel-v/4404bfceff16a6f8c68a9a295474fadd

My expectation is that whenever I add a new value to the stream, it will be displayed. Instead, nothing happens.

I found the source of the error. It's got to do with this line in async package: [StreamController.stream](https://github.com/dart-lang/sdk/blob/master/sdk/lib/async/stream_controller.dart#L459)

Together with AsyncPipes [identity check](https://github.com/dart-lang/angular2/blob/2.2.0/lib/src/common/pipes/async_pipe.dart#L72) things break. What happens is that the `obj` and `this._obj` will never be identical.

However, introducing a helper method would fix it.

```dart
  bool _maybeStreamIdentical(dynamic a, dynamic b) {
    if(a is Stream && b is Stream) {
      return identical(a, b) || a == b;
    } else {
      return identical(a, b);
    }
  }
```

zoechi:
What about changing

```dart
Stream<String> get stream => _ctrl.stream;
```
to
```dart
Stream<String> get _stream;
Stream<String> get stream => _stream ??= _ctrl.stream;
```

daniel-v:
That would work, I imagine. Either it has to be very well documented, that StreamControllers return a new Stream every time or make sure that it's not an issue. It's my 2 cent, that making sure that it's not an issue would space people some annoyment.

zoechi:
> that making sure that it's not an issue would spare people some annoyment.

I basically agree, but if the stream is not intended to be used this way, then I think it shouldn't be "fixed" in Angular, but in documentation.

johnpryan:
I'm trying to understand what the fix for this issue was. I have a `StreamController` with two separate streams created from it using `map`:

```dart
class FancyBloc {
  final StreamController<User> _sink;

  FancyBloc() : _sink = new StreamController.broadcast() {
  }

  Stream get firstName => _sink.stream.map((u) => u.firstName);
  Stream get lastName => _sink.stream.map((u) => u.lastName);

  ngOnInit() async {
    await new Future.delayed(new Duration(seconds: 1));
    _sink.add(new User()
      ..firstName = "Greg"
      ..lastName = "Gregorson");
  }
}

class User {
  String firstName;
  String lastName;
}
```

But in order to use asyncPipe, the result of calling `FancyBloc.firstName` and `FancyBloc.lastName` needs to return the same exact stream:

[...]

PR #261, opened 2017-01-31
Async pipe can now handle StreamController.stream's properly

daniel-v:
fixes dart-lang/angular2#260

I imagine a couple of test cases could be added to this fix.

### DIFF
--- a/lib/src/common/pipes/async_pipe.dart
+++ b/lib/src/common/pipes/async_pipe.dart
@@ -69,7 +69,8 @@ class AsyncPipe implements OnDestroy {
       this._latestReturnedValue = this._latestValue;
       return this._latestValue;
     }
-    if (!identical(obj, this._obj)) {
+    print('${_maybeStreamIdentical(obj, this._obj)}');
+    if(!_maybeStreamIdentical(obj, this._obj)) {
       this._dispose();
       return this.transform(obj);
     }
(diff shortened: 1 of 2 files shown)


lib/src/common/pipes/async_pipe.dart
```diff
@@ -69,7 +69,8 @@ class AsyncPipe implements OnDestroy {
       this._latestReturnedValue = this._latestValue;
       return this._latestValue;
     }
-    if (!identical(obj, this._obj)) {
+    print('${_maybeStreamIdentical(obj, this._obj)}');
```
matanlurey: Remove print.

lib/src/common/pipes/async_pipe.dart
```diff
@@ -69,7 +69,8 @@ class AsyncPipe implements OnDestroy {
...
+    if(!_maybeStreamIdentical(obj, this._obj)) {
```
matanlurey: Add a one/two line comment on why this is necessary for future devs.

lib/src/common/pipes/async_pipe.dart
```diff
@@ -114,4 +115,12 @@ class AsyncPipe implements OnDestroy {
       this._ref.markForCheck();
     }
   }
+
+  static bool _maybeStreamIdentical(dynamic a, dynamic b) {
```
matanlurey: Can be a little cleaner:

```dart
static bool _maybeStreamIdentical(a, b) {
  if (!identical(a, b)) {
    return a is Stream && b is Stream && a == b;
  }
  return true;
}
```

daniel-v:
Here is an example that did not work for me before: https://gist.github.com/daniel-v/4404bfceff16a6f8c68a9a295474fadd

There is a little more info in #260 .

daniel-v:
Could I get an update on this?

matanlurey:
Reminding myself to get this in tomorrow AM. Sorry about that.

PR #274, opened 2017-02-11
ApplicationRef.bootstrap: drop reference to TS code

chalin:
This isn’t the usual way that apps are bootstrapped, so just drop the TS code reference.

cc @kwalrath


kwalrath:
LGTM. Should we clarify further, @TedSander or @matanlurey?

PR #283, opened 2017-02-16
API docs: router CanDeactivate - add example

chalin:
Contributes to https://github.com/dart-lang/site-webdev/issues/367

cc @kwalrath

### DIFF
--- a/lib/src/router/interfaces.dart
+++ b/lib/src/router/interfaces.dart
@@ -96,21 +96,29 @@ abstract class CanReuse {
 /// if a component can be removed as part of a navigation.
 ///
 /// The `routerCanDeactivate` hook is called with two [ComponentInstruction]s as parameters,
-/// the
-/// first representing the current route being navigated to, and the second parameter
+/// the first representing the current route being navigated to, and the second parameter
 /// representing the previous route.
 ///
 /// If `routerCanDeactivate` returns or resolves to `false`, the navigation is cancelled. If it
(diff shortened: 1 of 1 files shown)


lib/src/router/interfaces.dart
```diff
@@ -96,21 +96,29 @@ abstract class CanReuse {
...
+/// @override
+/// /*FutureOr<bool>*/ routerCanDeactivate(next, prev) =>
+///     crisis == null || crisis.name == name
+///         ? true
+///         : _dialogService.confirm('Discard changes?');
 /// ```
+///
+/// See the [router documentation][router] for details.
+///
+/// [router]: docs/guide/router/5#candeactivate
```
kwalrath: I was thinking we might want to use absolute URLs for these, so that people reading the source code can more easily visit the links.

chalin: The link is correct, and the leading `docs/...` is a trigger for use of `https://webdev.dartlang.org/angular`.

There are pros and cons. Pros:
- The docs have been moving around so the `docs` alias gives us one level of indirection.
- I.e., if the docs site moves, then we only need to change the path to the `docs` in one place.
- It is far shorter to type :).

Cons:
- It is "magic"
- Other tools can't follow the link.

Its your call.

If this is the only issue, can you accept this PR and then we can make a sweeping change? (If you could create an issue for that.)

PR #315, opened 2017-03-17
Add a warning for Dartium/unchecked.

matanlurey:
Closes https://github.com/dart-lang/angular2/issues/258

### DIFF
--- a/lib/src/core/application_ref.dart
+++ b/lib/src/core/application_ref.dart
@@ -33,6 +33,17 @@ PlatformRefImpl createPlatform(Injector injector) {
     }
     return true;
   });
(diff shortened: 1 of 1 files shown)


lib/src/core/application_ref.dart
```diff
@@ -33,6 +33,17 @@ PlatformRefImpl createPlatform(Injector injector) {
     }
     return true;
   });
+  // Dartium. Otherwise will be tree-shaken out.
+  if (identical(1, 1.0)) {
```
kevmoo: src/facade/lang.dart already has some of these checks...might be useful...

PR #317, opened 2017-03-17
Update NgModel API docs

### LINKED ISSUE #264: NgModel API doc has bad link to the live example
A formatting problem in https://webdev.dartlang.org/angular/api/angular2.common/NgModel-class makes what's supposed to be live link look like this:

image

kwalrath:
Fixes #264


kwalrath:
Reopening because this'll get automatically closed when we sync.

PR #320, opened 2017-03-28
Enable new travis hotness

kevmoo:


matanlurey:
I trust you that this works assuming travis passes. We are still merging internally first, right?

kevmoo:
> I trust you that this works assuming travis passes. We are still merging internally first, right?

yup yup

PR #323, opened 2017-03-29
Sync to Google HEAD

matanlurey:


pubspec.yaml
```diff
@@ -1,9 +1,9 @@
 name: angular2
-version: 3.0.0-alpha+1
+version: 3.0.0-beta
```
kevmoo: Need to update `angular_test` to support beta

@matanlurey

matanlurey: Yep, going to tackle this today.

matanlurey:
Going to merge and then start a new commit to fix the breakage.

PR #336, opened 2017-04-06
Use uri instead of uriContent for sorting UriBasedDirective

kevmoo:
The former seems to be less likely null

Fixes https://github.com/dart-lang/angular2/issues/302


alorenzen:
This was already fixed internally by Ferhat. The issue was actually in the cascade call add(imports..sort(_compareUri)) on line 103 and 104 of code_builder.dart

kevmoo:
Doing this on the inside...

PR #338, opened 2017-04-06
API docs: Router OnActivate

chalin:
cc @kwalrath @kevmoo

### DIFF
--- a/lib/src/router/interfaces.dart
+++ b/lib/src/router/interfaces.dart
@@ -16,8 +16,16 @@ import 'instruction.dart' show ComponentInstruction;
 ///
 /// ### Example
 /// ```dart
-/// // {@disabled-source "router/ts/on_activate/on_activate_example.ts" region="routerOnActivate"}
+/// // {@source "docs/router/lib/crisis_center/crisis_detail_component.dart" region="routerOnActivate"}
+/// @override
+/// void routerOnActivate(next, prev) {
+///   print('Activating ${next.routeName} ${next.urlPath}');
+/// }
 /// ```
+///
+/// See the [router documentation][router] for details.
+///
+/// [router]: https://webdev.dartlang.org/angular/guide/router/5#onactivate
 abstract class OnActivate {
   dynamic /* dynamic | Future< dynamic > */ routerOnActivate(
       ComponentInstruction nextInstruction,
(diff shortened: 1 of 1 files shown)


lib/src/router/interfaces.dart
```diff
@@ -16,8 +16,16 @@ import 'instruction.dart' show ComponentInstruction;
...
+/// // {@source "docs/router/lib/crisis_center/crisis_detail_component.dart" region="routerOnActivate"}
+/// @override
+/// void routerOnActivate(next, prev) {
+///   print('Activating ${next.routeName} ${next.urlPath}');
+/// }
 /// ```
+///
+/// See the [router documentation][router] for details.
+///
+/// [router]: docs/guide/router/5#onactivate
```
kwalrath: Not necessary now, but for the future it'd be nice to have a real URL in here, so it's easier to (1) confirm the link if you're reviewing and (2) follow the link if you're reading source code instead of rendered docs.

chalin: Agreed. Created https://github.com/dart-lang/site-webdev/issues/534 to track this.

chalin: Fix URL for this file.

chalin:
Haven't given it any further thought. Do we have an issue open for that? I believe that you'd like this to be hidden in the generated dartdocs? It actually seems like useful information to have. We might instead consider converting it into a link into the GH repo source file. WDYT?

kwalrath:
I think that, as a rule, we should omit the `@source` comment from the generated docs. Having the first line not be the actual example seems like a bad user experience. (The reader can look at the doc comment to get this information, or the doc can explicitly link to the example.)

A possible tweak is to add a styled link that's not shown, by default, so that the doc website can display the info or not.

cc @jcollins-g

chalin:
> add a styled link that's not shown, by default, so that the doc website can display the info or not.

SGTM

chalin:
We could handle it in dartdoc, or I could write a filter which would be applied when copying over the ng2 API docs to webdev. It might be preferable to experiment with the latter first. WDYT?

kwalrath:
Experimenting with the latter SGTM.

chalin:
@kwalrath - followup issue is #551

PR #359, opened 2017-04-25
Cheatsheet updates: SwitchCase, EventEmitter

chalin:
- Dropped `EventEmitter`
- `ngSwitchWhen` --> `ngSwitchCase`

cc @kwalrath

### DIFF
--- a/doc/cheatsheet/built-in-directives.md
+++ b/doc/cheatsheet/built-in-directives.md
@@ -27,16 +27,16 @@ See: [Template Syntax](/angular/guide/template-syntax),
 @cheatsheetItem
 syntax:
 `<div [ngSwitch]="conditionExpression">
-  <template [ngSwitchWhen]="case1Exp">...</template>
-  <template ngSwitchWhen="case2LiteralString">...</template>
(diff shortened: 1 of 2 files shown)


chalin:
Yes. Here:
screen shot 2017-04-25 at 1 57 38 pm

And here:
screen shot 2017-04-25 at 1 57 26 pm

kwalrath:
Looks great. @kevmoo can you incorporate this?

chalin:
@kevmoo, @ferhatb - after some discussion, @kwalrath and I agreed to move the cheatsheet doc sources to site-webdev ... so that you won't need to be bothered anymore with updates like these in the future.

chalin:
See #360

zoechi:
shouldn't `sync: true` be suggested for the `StreamController`?

matanlurey:
Synchronous stream controllers are usually only meant to adapt already
asynchronous APIs that aren't Stream based (similar to Completer is to
Future).

zoechi:
I haven't checked for a long time. I just thought to remember that this was made the default in EventEmitter long ago because of timing issues.
If this isn't the case please ignore my comments.

PR #365, opened 2017-04-29
Expose TouchFunction and ChangeFunction from ControlValueAccessor

creisman:
Expose TouchFunction and ChangeFunction from ControlValueAccessor to prevent needing to import from src/.


creisman:
Apparently MOE will generate this automatically. Closing.

PR #372, opened 2017-05-10
Sync to Google's HEAD

matanlurey:
/cc @kevmoo (who's out) and @filiph 

I don't think it makes sense to make a point release until @alorenzen's commit to add the analyzer-based compiler via a transformer flag lands. But this gets us closer.


lib/src/compiler/runtime_compiler.dart
```diff
@@ -38,10 +38,10 @@ class RuntimeCompiler implements ComponentResolver {
   final StyleCompiler _styleCompiler;
   final ViewCompiler _viewCompiler;
   XHR _xhr;
-  Map<String, Future<String>> _styleCache = new Map<String, Future<String>>();
-  var _hostCacheKeys = new Map<Type, dynamic>();
-  var _compiledTemplateCache = new Map<dynamic, CompiledTemplate>();
-  var _compiledTemplateDone = new Map<dynamic, Future<CompiledTemplate>>();
+  final _styleCache = new Map<String, Future<String>>();
```
zoechi: isn't using collection literals the preferred way in Dart? http://dart-lang.github.io/linter/lints/prefer_collection_literals.html

matanlurey: Honestly given that this is soon to be dead code it's sort of like a P5.

matanlurey:
Plan on fixing travis in a future PR.

PR #376, opened 2017-05-12
Sync to Google's HEAD.

matanlurey:
Assuming travis is green this will be a prospective release point.


matanlurey:
Still fixing.

PR #377, opened 2017-05-12
Sync to Google's HEAD

matanlurey:


matanlurey:
Expecting a failure still, more commits on-the-way.

PR #379, opened 2017-05-12
Sync to Google HEAD

matanlurey:


matanlurey:
This will be green, so submitting :)

PR #391, opened 2017-05-16
Sync to HEAD.

matanlurey:


matanlurey:
Assuming the rest of the build will pass at this point.

PR #396, opened 2017-05-18
API doc examples: minor misc updates

chalin:
cc @kwalrath


kwalrath:
@matanlurey can we get this into 3.1.0?

PR #477, opened 2017-06-29
Clean up compile_metadata.dart

alorenzen:
1) Remove toJson/fromJson. This was only used in the old transformers.
2) Remove emitPrefix, which is always true now.


matanlurey:
Do not approve on GitHub. We don't merge here.

kevmoo:
What's the story here? @matanlurey @alorenzen @leonsenft ?

matanlurey:
This landed but wasn't marked. Closed.

alorenzen:
This was out for review internally, but hasn't actually landed yet. This should be left open until it is actually merged into master, no?

kevmoo:
True!

PR #487, opened 2017-06-30
Cache pub-cache, only build master branch

kevmoo:
Avoid double-building branches and pull requests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,11 @@ dart:
   - dev
   - stable
 script: ./tool/travis.sh
+
+# Only building master means that we don't run two builds for each pull request.
+branches:
+  only: [master]
+
+cache:
+ directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,3 +4,11 @@ dart:
   - dev
   - stable
 script: ./tool/travis.sh
+
+# Only building master means that we don't run two builds for each pull request.
+branches:
+  only: [master]
```
matanlurey: Do we want `github-sync` too?

kevmoo: Nope. We always create a PR, right?

matanlurey: Ack.

kevmoo: We can always add it back – although it just doubles the runtime if it's going to end up being a PR, too.

PR #490, opened 2017-06-30
Separate tests into shards

kevmoo:


kevmoo:
@matanlurey PTAL – this also enables dartfmt tests

PR #512, opened 2017-07-12
Sync fixes for 4.0.0-alpha

matanlurey:


matanlurey:
Only failure is a golden file flake.

PR #518, opened 2017-07-13
GitHub sync

alorenzen:


matanlurey:
@alorenzen We should probably blacklist the source_gen golden tests :-/

alorenzen:
This actually looks like a legit failure to me. The golden files are not actually importing angular/angular.template.dart.

alorenzen:
OK, I found the issue with the golden files. When we made usePlaceholder the default again, we didn't change it for the generate.dart script (literally the one place that it's intended to not use the default :stuck_out_tongue: )

It's an easy enough fix, so we can include it in this PR.

matanlurey:
Great!

matanlurey:
@alorenzen We need to ignore dartfmt for dev, otherwise LGTM.

PR #520, opened 2017-07-14
Only run dartfmt on stable.

alorenzen:


alorenzen:
Verified that this only outputs "Any unformatted files?" for stable, but not dev.

alorenzen:
Merged in #518

PR #524, opened 2017-07-15
Cleanup to Differ classes

kevmoo:


kevmoo:
https://critique.corp.google.com/#review/162092687

PR #525, opened 2017-07-15
Use @MirrorsUsed(..) annotiation

Xexanos:
This commit introduces the @MirrorsUsed(..) annotation as described here: https://api.dartlang.org/stable/1.24.2/dart-mirrors/MirrorsUsed-class.html
This leads to less generated code:
Before: `Compiled 6,085,576 characters Dart to 9,543,914 characters JavaScript in 19.4 seconds`
After: `Compiled 6,085,628 characters Dart to 787,162 characters JavaScript in 4.60 seconds`

### DIFF
--- a/angular/lib/src/core/reflection/reflection_capabilities.dart
+++ b/angular/lib/src/core/reflection/reflection_capabilities.dart
@@ -1,3 +1,4 @@
+@MirrorsUsed(symbols: 'classMirror')
 import 'dart:mirrors';
 
 import 'package:angular/src/core/metadata/lifecycle_hooks.dart';


zoechi:
Did you use the `reflection_remover` transformer when you tried without `@MirrorsUsed`?

```
- angular2/transform/reflection_remover:
    $include:
      - web/main.dart
```

Xexanos:
No, I didn't, but tried it out.
I stumbled across this phenomenon, when I was using `dart2js` to compile
the `main.dart` of my project using `dart-lang/angular`. Compiling it
produced a warning which lead me here.
Is using that transformer best practice? I just liked the idea of using
a simple annotation directly in `dart-lang/angular` and was wondering
why it is not used.

zoechi:
Reflection (`dart:mirrors`) breaks tree-shaking because this way dart2js can't know anymore which parts of the code are actually used and which not, this is why your build output becomes so big.

`@MirrorsUsed()` allows to guide dart2js tree-shaking when reflection is used.
The `reflection_remover` transformer replaces `dart:mirrors` usage by generates code that avoids reflection entirely.

AFAIK this is way better than guiding dart2js.

matanlurey:
There is no more reflection (at all) in AngularDart 4.0.

PR #550, opened 2017-07-20
WIP: Use 1.24 Function syntax

kevmoo:
Blocked on a new analyzer release - https://github.com/dart-lang/sdk/issues/30239

Using alpha pkg/analyzer to see if travis gets happy
Looks like we need to update to use the 1.25.0-dev SDK, too!

### DIFF
--- a/angular/lib/src/core/linker/app_view.dart
+++ b/angular/lib/src/core/linker/app_view.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 import 'dart:html';
 import 'dart:js_util' as js_util;
 
-import 'package:func/func.dart';
 import 'package:meta/meta.dart';
 import 'package:angular/src/core/change_detection/change_detection.dart'
     show ChangeDetectorRef, ChangeDetectionStrategy, ChangeDetectorState;
@@ -557,7 +556,7 @@ abstract class AppView<T> {
     domRootRendererIsDirty = true;
   }
 
-  VoidFunc1<E> eventHandler0<E>(VoidFunc0 handler) {
+  void Function(E) eventHandler0<E>(void Function() handler) {
     return (E event) {
       markPathToRootAsCheckOnce();
       if (NgZone.isInAngularZone()) {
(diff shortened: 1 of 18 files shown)


angular/lib/src/core/linker/app_view.dart
```diff
@@ -584,7 +583,7 @@ abstract class AppView<T> {
     domRootRendererIsDirty = true;
   }
 
-  VoidFunc1<E> eventHandler0<E>(VoidFunc0 handler) {
+  void Function(E) eventHandler0<E>(void Function() handler) {
```
leonsenft: Are we decided on what syntax to use when declaring function parameters? Both are valid:
```dart
void Function(E) eventHandler0<E>(void Function() handler) {}

void Function(E) eventHandler0<E>(void handler()) {}
```
I'm in favor of the latter as it's more concise, but I can understand the desire to be consistent with the former.

matanlurey: I think we decided always the former, the latter may be removed in a future version of Dart.

kevmoo: And encouraging the new syntax is goodness.

```dart
// No error, but this handle has no argument type
foo(void handler(int)) {};

// Does what you expect
foo(void Function(int)) {};
```

leonsenft:
LTGM, although some of the files still contain function typdefs. It won't let me comment on them directly but see `*_control_acccesors.dart` for `void registerOnChange(ChangeFunction fn)`.

kevmoo:
> LTGM, although some of the files still contain function typdefs. It won't let me comment on them directly but see *_control_acccesors.dart for void registerOnChange(ChangeFunction fn).

Yeah, those are used across many files – left them for now until we decide what to do.

angular/lib/src/core/linker/app_view.dart
leonsenft: In that case there are some old style function parameters in these files you could update as well.

kevmoo: Yeah...but that's hard to regexp

I'd almost want a lint for that

PR #561, opened 2017-07-27
GitHub sync

alorenzen:


matanlurey:
@alorenzen It looks like we need the `source_gen` version bumped.

PR #567, opened 2017-08-01
stop using deprecated cli_util method

kevmoo:


kevmoo:
Fixed in 3968e00a5810a88c6b656f732e5368daa40d55a6

PR #591, opened 2017-08-24
GitHub sync

alorenzen:


angular/CHANGELOG.md
```diff
@@ -1,4 +1,20 @@
...
+### Breaking changes
+
+* `@Pipe`-annotated classes are no longer considered `@Injectable`, in that they
+  aren't usable within a `ReflectiveInjector`. You can get this behavior back by
+  adding the `@Injectable()` annotation to the `@Pipe`-annotated class. Similar
+  changes are in progress for `@Component` and `@Directive`.
+
+* `PLATFORM_{PIPES|DIRECTIVES|PROVIDERS}`, which was only supported in an older
+  version of the compiler, was removed. All of these must be manually included
+  in lists in an `@Directive` or `@Component` annotation.
```
zoechi: I find this paragraph unclear. What does "All of these" mean exactly when `PLATFORM_{PIPES|DIRECTIVES|PROVIDERS}` was removed?
Should the used directives be added individually, or are `PLATFORM_{PIPES|DIRECTIVES|PROVIDERS}` still available but must be added to `directives: [...]` instead of to transformer options (but that was already removed in Angular 3 AFAIR)

kwalrath: Clarifying would be good. We might also want to consider linking to the doc changelog: https://webdev.dartlang.org/angular/changelog.

PR #605, opened 2017-08-30
Fix pubspec for analyzer_plugin

kevmoo:
Ensure packages are resolved before analyzing


kevmoo:
CC @MichaelRFairhurst 

...it couldn't be this easy, could it?

MichaelRFairhurst:
This should be everything you need, yes!

PR #624, opened 2017-09-11
Optimization for Barback: Don't wait for outputs.

matanlurey:
/cc @chalin if you want to see if this improves build-times for the examples.

This is based on a non-exhaustive theory that waiting for outputs in Barback is the source of significant (blocking) time, and can be optimized away in most cases where `$exclude` and `$include` are not used.

In the cases they are, a new optional flag, `generator_inputs`, is available. We could optionally flip this as well and support something like a `not_generator_inputs`, if that is more useful. 

**FOR FEEDBACK ONLY: DO NOT MERGE**

### DIFF
--- a/_tests/test/common/forms/integration_test.dart
+++ b/_tests/test/common/forms/integration_test.dart
@@ -872,7 +872,9 @@ class InputWithoutTypeTest {
 }
 
 @Component(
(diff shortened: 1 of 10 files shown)


alorenzen:
IIRC, this as been submitted elsewhere. Can it be closed now?

matanlurey:
Yes, this can be closed.

PR #666, opened 2017-10-11
Fix the generator_inputs optimization.

matanlurey:
**DO NOT SUBMIT**: For testing only.

If it improves builds (it does for me locally), we'll release a 4.1 and update `angular_compiler`.

To try it, clone/fetch this PR, and add the following:

```yaml
dependency_overrides:
  angular:
    path: ../angular/angular
  angular_compiler:
    path: ../angular/angular_compiler
```

When running `pub serve -v` you should see something like:

`Avoiding a lookup on angular|lib/src/core/security.dart (matched generator_inputs)`

... many many times.

/cc @kevmoo


kevmoo:
No speed improvement as far as I can see...

matanlurey:
@kevmoo Could you share your `pub serve -v`?

matanlurey:
Abandoned for now.

PR #673, opened 2017-10-18
github-sync

alorenzen:
Invokes functional directives after constructing host element children


matanlurey:
I'm going to merge on red since we haven't merged in ~2 weeks.

We can then work to get the repository green, but it might require lots of experimentation - and it will be easier to test via PRs than doing the whole google3 -> copybara -> PR.

In reality it shouldn't matter since we don't plan on releasing anytime soon :)

PR #685, opened 2017-11-07
Verify travis fixes

alorenzen:
** DO NOT MERGE** 
This is just to verify travis fixes.


kevmoo:
Can this PR be killed?

alorenzen:
When the tests are running properly with build_runner on travis, then yes. But we are not there yet...

kevmoo:
can we drop the -dev.6.0 platform? Will make things run faster...

alorenzen:
This is just a PR for hacking, not something we intent to land. I really don't care how long it takes, as long as it actually gives us a useful signal.
I'm not really keeping this in a pristine state. Just throwing something out there for Jake to hack on...

PR #701, opened 2017-11-21
use a non-git dep for bazel_worker

jakemac53:
This should work now with the latest version of bazel_worker  (0.1.7) and the additional environment variable which already existed.

### DIFF
--- a/_tests/tool/travis.sh
+++ b/_tests/tool/travis.sh
@@ -10,8 +10,6 @@ set -e
 dartanalyzer --fatal-warnings .
 
 pub run test -p vm -x codegen
-
-export BUILD_MAX_WORKERS_PER_TASK=1
 dart tool/build.dart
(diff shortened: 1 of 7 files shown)


_tests/tool/travis.sh
```diff
@@ -10,6 +10,8 @@ set -e
 dartanalyzer --fatal-warnings .
 
 pub run test -p vm -x codegen
+
+export BUILD_MAX_WORKERS_PER_TASK=1
```
alorenzen: Will this work if we just define it at the top-level `travis.sh`? 

That way, it would apply to all of the sub-packages that use `build_runner`.

jakemac53: oh it should, I will try adding that

alorenzen:
This was submitted as https://github.com/dart-lang/angular/commit/ebcc393e05649872cb3ccb612426f610412b824c

PR #708, opened 2017-11-24
[angular_test] Support use of external pub serve

### LINKED ISSUE #469: Add a --pub-serve argument
_From @matanlurey on June 19, 2017 16:30_

This would disable launching `pub serve` by default, and instead delegate to an instance:

```bash
$ pub serve --port 1234
```

```bash
$ angular_test --pub-serve=1234
```

This is important because some "transformers" like dartdevc take quite a bit on first-use, and will often exceed `pkg/test`'s timeouts. Only running as a daemon (incremental rebuilds) makes it worth it, and adding daemon support for `angular_test` is quite a bit of work to get right.

_Copied from original issue: dart-lang/angular_test#74_

chalin:
Support is added w/o creating new command line flags. To have angular_test use an external pub serve (rather than launch its own) use "none" as the value of `--experimental-serve-script`; specify the external server port using `--port`. E.g.:

```shell
pub run angular_test --experimental-serve-script=none --port=8081 ...
```

Fixes #469

### DIFF
--- a/angular_test/pubspec.yaml
+++ b/angular_test/pubspec.yaml
@@ -2,7 +2,7 @@ name: angular_test
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/angular
 description: Testing runner and library for AngularDart
-version: 1.0.2-alpha+2-dev
+version: 1.0.2-alpha+1
 
 environment:
   sdk: '>=2.0.0-dev.3.0 <2.0.0'
--- a/angular_test/pubspec.yaml
+++ b/angular_test/pubspec.yaml
@@ -2,7 +2,7 @@ name: angular_test
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/angular
 description: Testing runner and library for AngularDart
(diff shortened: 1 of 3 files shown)


matanlurey:
Looks reasonable. I'm back in the office monday and can merge.

angular_test/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: angular_test
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/angular
 description: Testing runner and library for AngularDart
-version: 1.0.2-alpha+1
+version: 1.0.2-alpha+2-dev
```
alorenzen: Please don't bump the versions yourself. We will handle this when we go to publish, based on all of the changes that are included since the last release.

Instead, just add an entry in the CHANGELOG.

chalin: Ok.

chalin: Done

matanlurey:
Merging as-is @alorenzen

PR #834, opened 2018-02-05
Add markAsUntouched method to AbstractControl

elvisun:
There is currently no way of restoring a control's state once it's been touched. So it's impossible to suppress the error message, which could be useful in some cases. For example, [delaying error message until input blur](https://embed.plnkr.co/04anApG7LpPz6pTfCOHD/). 

There's the same [APIs in typescript version of angular forms](https://github.com/angular/angular/blob/master/packages/forms/src/model.ts#L291), I think we should have the same in dart.

### DIFF
--- a/angular_forms/lib/src/model.dart
+++ b/angular_forms/lib/src/model.dart
@@ -75,6 +75,10 @@ abstract class AbstractControl {
   void markAsTouched() {
     _touched = true;
   }
+   
+  void markAsUntouched() {
+    _touched = false;    
+  }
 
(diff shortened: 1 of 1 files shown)


matanlurey:
@TedSander, care to review? (Or we can wait for @alorenzen)

matanlurey:
@elvisun Looks good. Do you mind adding a minimal test for this feature?

matanlurey:
I'm closing due to lack of activity. Please feel free to re-open.

PR #837, opened 2018-02-06
chore(Travis): Try removing --low-resources-mode

matanlurey:
GCE-based containers (we use one now) have ~8Gb of Ram and 2 (Burstable) Cores:
https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

We should see if we can remove `--low-resources-mode`, if possible.

### DIFF
--- a/angular/lib/src/di/providers.dart
+++ b/angular/lib/src/di/providers.dart
@@ -6,8 +6,6 @@ import 'package:meta/meta.dart';
 
 import '../core/di/opaque_token.dart';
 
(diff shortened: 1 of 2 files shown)


matanlurey:
_Looks_ like it works just fine. Will let @jakemac53 and @natebosch give the yay/nay.

(One big difference now is we never run `pub run test` concurrently with the build system)

jakemac53:
LGTM! that is much faster now haha

PR #860, opened 2018-02-08
Make main.dart.js load upon subdirectory refresh

exo-pla-net:
Without a leading "/", main.dart.js won't load on site.com/sub/directories. (The server will look for site.com/sub/directories/main.dart.js, and upon not finding such a file, it will rewrite main.dart.js to *this* index.html file, causing browser console error "Unexpected token <"  whenever a subdirectory is refreshed.)


exo-pla-net:
I signed it!

matanlurey:
@leonsenft

matanlurey:
It looks like this example is out of date, and we're actually going to move it to `examples`.

Thanks for the pull-request, we'll make sure to have the href setup properly there.

PR #905, opened 2018-02-16
github sync

alorenzen:
refactor(Forms): Use new provider and token features in angular_forms.


matanlurey:
Merging on red so we can fix our export scripts.

PR #912, opened 2018-02-21
github sync

alorenzen:
feat(Compiler): Add a testing apparatus for the compiler.

Interesting parts:
* _tests/lib/compiler.dart: Reusable functions for tests.
* _tests/test/compiler_integration/example_test.dart: Example test.

... most of the rest is minor refactoring to avoid circular dependencies between cli.dart and compiler.dart (for Bazel), and I fixed a bug in the compiler (and tested that fix in example_test.dart).


matanlurey:
@alorenzen I'm merging on red (we are already broken on MASTER) so it will be easier to debug/figure out how to fix this.

PR #929, opened 2018-02-26
github sync

alorenzen:
Add a type annotation, T, to AbstractControl and ControlGroup, which is tied to the StreamController parameter for [_valueChanges] and the Stream parameter for [valueChanges].

This allows users of ControlGroup to listen on [valueChanges] and expect a certain type in their listen callback.


matanlurey:
Merging on red.

PR #959, opened 2018-02-27
Start XVFB prior to testing

grouma:
This was removed when we were running with Chrome Headless. Unfortunately Headless has some issues with the latest SDK https://github.com/dart-lang/test/issues/772 so we must run with full Chrome for the time being. This will allow us to run with the latest test package and latest SDK without issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -166,3 +166,9 @@ cache:
     - examples/hello_world/.dart_tool
     - examples/hello_world_no_reflector/.dart_tool
     # END MANUAL EDITS.
+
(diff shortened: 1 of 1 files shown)


alorenzen:
Is there any reason why a dart4web repository wouldn't want this config? Does it make sense to add it to the standard dart travis environment instead?

natebosch:
> Does it make sense to add it to the standard dart travis environment instead?

The standard dart travis tasks already handles this. We aren't using the standard tasks because we aren't a single-package repo.

Issue #939, opened 2018-02-26
hacker_news_pwa: Publish on Firebase/Github.io


matanlurey:
Just fun :)

matanlurey:
I'd like to set this up via travis, but we aren't green right now.

For now I'll do a single manual push.

matanlurey:
Done as https://dart-lang.github.io/angular/ for now.

... can find a better home later on :)

filiph:
That's pretty awesome.

![screen shot 2018-02-26 at 19 50 15](https://user-images.githubusercontent.com/919717/36710104-c249195a-1b30-11e8-88d3-1920a79407b0.png)

The page loads & paints & gets interactive in 2780ms, even on the simulated slow connection. That's really good for something like AngularDart, which doesn't even try to optimize for microbenchmarks like this. Just look at https://hnpwa.com/, the latest React score is 91 and TTI is 2570ms, AngularJS gets 4-6 seconds, etc.

Ignoring SEO (that's another discussion altogether), there are only a couple of issues. 

1. We lose some perf points by not caching `angular/main.dart.js` and `images/logo.svg` for long enough. I don't think there's anything we can do on github.io, but on Firebase Hosting, you just increase TTL to something like 24 hours and Lighthouse is happy (because that's when ISPs start to cache things, which makes things much faster for the end user). So: impossible fix on github.io, easy fix on firebase hosting.

2. The rest of points lost is "perceptual speed index". This is best explained [here](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/metrics/speed-index). Basically, the earlier we can show the UI (even if it's only a placeholder) the better. Maybe show the top bar in the static HTML? Even better, "make room" for the loaded articles with white divs? That alone makes a lot of the initial pixels correspond to the final result.

    And this is not just about "beating a score", by the way. The apps that do this kind of optimization _do_ perceptually load faster. It's a very different experience looking at a gray screen for 2 seconds than at the beginnings of a UI.

Congrats to all!

In summary, consider:

* Moving this to Firebase Hosting and cache the two files more aggressively.
* [Submitting](https://github.com/tastejs/hacker-news-pwas/blob/master/CONTRIBUTING.md) the app to https://hnpwa.com/.
* Adding the most visually prominent parts of the UI (app bar, white area with list of articles) to the static HTML + CSS.

matanlurey:
[...]

PR #965, opened 2018-02-27
chore(Example): Add 10 minutes of caching for firebase hosting.

matanlurey:
See https://github.com/dart-lang/angular/issues/939.


examples/hacker_news_pwa/firebase.json
```diff
@@ -15,8 +15,14 @@
...
+        "headers": [{
+          "key": "Link",
+          "value": "</main.dart.js>;rel=preload;as=script"
+        }, {
+          "key" : "Cache-Control",
+          "value" : "max-age=86400"
+        }]
       }
-    ]
+    ],
```
kevmoo: Will trailing common break JSON parsers? Tis not Dart code 😄

matanlurey: Good catch, done.

Issue #969, opened 2018-02-28
HN example: Increase firebase cache times for important assets


kevmoo:
See https://www.webpagetest.org/performance_optimization.php?test=180228_Y9_308b0df83539fbb26d30a51ea3067c7b&run=3#cache_static_content

Add 24-hour expire for `pwa.dart.js`, `images/logo.svg`

...and `main.dart.js` ?

matanlurey:
As far as I know I did?

https://github.com/dart-lang/angular/blob/7024dbe70531f5b8b3450992dfcd957d2e798e3b/examples/hacker_news_pwa/firebase.json#L15-L26

@kevmoo

leonsenft:
Is there someway to indicate a request isn't a static asset? The 80/100 score for using CDN on static assets blames a REST API we hit, which isn't a static asset.

matanlurey:
I think we did this. Feel free to re-open if I'm mistaken @kevmoo.

kevmoo:
You did it for the root file, but not the other stuff...

I'll dig in...

PR #973, opened 2018-02-28
Set expire time for js and svg at 1 day - 86400

kevmoo:
Fixes https://github.com/dart-lang/angular/issues/969


kevmoo:
I did a deploy and confirmed...

kevmoo:
https://www.webpagetest.org/performance_optimization.php?test=180228_HJ_748430d935960e20161d436a05c9d494&run=3#cache_static_content

We go from an F to a C grade with this.

7 days is not enough. Needs to be 7 days + 1s. Weird...

Issue #960, opened 2018-02-27
Fix SEO in HN app


kevmoo:
https://www.webpagetest.org/lighthouse.php?test=180228_Y9_308b0df83539fbb26d30a51ea3067c7b&run=3#seo

![screen shot 2018-02-27 at 1 35 42 pm](https://user-images.githubusercontent.com/17034/36756174-1fdab528-1bc3-11e8-8dc3-de120aa077d3.png)

kevmoo:
Fixed in https://github.com/dart-lang/angular/commit/767d578c9139c14a5542ae2f02843f69b7d284b7

PR #975, opened 2018-02-28
Fix SEO issues from Light House

kevmoo:
Fixes https://github.com/dart-lang/angular/issues/960

### DIFF
--- a/examples/hacker_news_pwa/web/index.html
+++ b/examples/hacker_news_pwa/web/index.html
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <meta name="theme-color" content="#212121">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <meta name="Description" content="An implementation of a Hacker News readers as a Progressive Web Apps using the AngularDart framework.">
+  <meta name="Description" content="A Hacker News reader implemented as a Progressive Web App using AngularDart.">
   <base href="/">
   <title>Hacker News - AngularDart PWA</title>
   <script defer src="main.dart.js"></script>
--- a/examples/hacker_news_pwa/web/index.html
+++ b/examples/hacker_news_pwa/web/index.html
@@ -4,6 +4,7 @@
   <meta charset="utf-8">
   <meta name="theme-color" content="#212121">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="Description" content="An implementation of a Hacker News readers as a Progressive Web Apps using the AngularDart framework.">
   <base href="/">
   <title>Hacker News - AngularDart PWA</title>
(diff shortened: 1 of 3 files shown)


examples/hacker_news_pwa/web/index.html
```diff
@@ -4,6 +4,7 @@
   <meta charset="utf-8">
   <meta name="theme-color" content="#212121">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="Description" content="An implementation of a Hacker News readers as a Progressive Web Apps using the AngularDart framework.">
```
leonsenft: Suggestion:
"A Hacker News reader implemented as a Progressive Web App using AngularDart."

Otherwise, if you'd prefer to keep the original description, "readers" and "Apps" shouldn't be plural.

kevmoo:
> Was this for testing? Can't you just use the Audit tab in Chrome during local development?

I was using local testing – but then I wanted to see if/how it changed the site score.

In the future, I will deploy to my own fork so we don't mess things up...

kevmoo:
Also: https://github.com/dart-lang/angular/issues/978

So we can track which commit is currently deployed...

leonsenft:
> I was using local testing – but then I wanted to see if/how it changed the site score.
>
> In the future, I will deploy to my own fork so we don't mess things up...

Not a big deal since this isn't a production app or anything, but as Matan mentioned in #978, there could be confusion as to what version is deployed. Also, now that it's in the repo, users may be visiting it so we shouldn't have experiments being pushed there just in case we break it (again not a huge deal at the moment since deferred loading is broken).

leonsenft:
Also, now that we've bumped up the font size, the UI looks pretty bad on mobile. There's a ton of inconsistent wrapping (there always could be if users had long names). We should maybe make a more intentional effort to place the author, timestamp and comments link in a way that renders more consistently.

kevmoo:
> Also, now that we've bumped up the font size, the UI looks pretty bad on mobile. There's a ton of inconsistent wrapping (there always could be if users had long names). We should maybe make a more intentional effort to place the author, timestamp and comments link in a way that renders more consistently.

ack...

kevmoo:
@leonsenft – should we fix that first before we merge this?

kevmoo:
Yup!

leonsenft:
> should we fix that first before we merge this?

No, there's so many UI improvements to make I wouldn't bother now. Let's focus on the scores to set a baseline. Then we can iterate on UI changes under the constraints of maintaining the score.

kevmoo:
SGTM

leonsenft:
I have no idea, I literally clicked the `Add your review` link where it says `Review required` at the bottom and nothing changed. It doesn't matter since we merged internally, so don't worry about it. LGTM.

PR #981, opened 2018-02-28
Use travis deploy config

kevmoo:
Fixes https://github.com/dart-lang/angular/issues/978

PR #987, opened 2018-03-01
refactor(Travis): Replace mono_repo with hand-written script.

matanlurey:
Closes https://github.com/dart-lang/angular/pull/981.


kevmoo:
Do we feel mono_repo won't "scale" to this setup?

kevmoo:
> Do we feel mono_repo won't "scale" to this setup?

Saw the internal discussion. Deleting.

dev/travis/postfix.yaml
```diff
@@ -0,0 +1,14 @@
+# Create a merged output build/.
+before_deploy:
+  - pushd examples/hacker_news_pwa
```
kevmoo: We've already built in a stage before, right? Why do it again?

...or just running it expecting a `no-opt`?

matanlurey: Right, expectation is a no-op, but I'm not sure deployment can see the build cache, so being extra cautious.

kevmoo: Well. It's weird.

`the -d example/hacker_new...` condition makes sure that directory exists.

So the deploy (and pre deploy) will only run in that case.

...I guess it should be a no-op, but would be good to validate.

matanlurey: Definitely. Let's watch the logs and see what happens.

kevmoo: Won't see anything here. Only when it lands in master.

Unless you want to push this to a branch here and enable it – to sanity check...

.travis.yml
```diff
@@ -1,174 +1,261 @@
...
+
+    - stage: presubmit
+      script: ./tool/travis.sh analyze
       env: PKG="examples/hacker_news_pwa"
-      dart: dev
-    - stage: build
-      script: ./tool/travis.sh command_7
+
+    - stage: building
+      script: ./tool/travis.sh build
```
kevmoo: should be `build:release`?

.travis.yml
```diff
@@ -1,174 +1,261 @@
...
+          - examples/hello_world_no_reflector/.dart_tool
 
-stages:
-  - analyze
-  - build
-  - unit_test
+# Create a merged output build/.
+before_deploy:
+  - pushd examples/hacker_news_pwa
+  - pub run build_runner build -o build --fail-on-severe
```
kevmoo: should have `--config release` ?

.travis.yml
kevmoo: @matanlurey ?

Also, the deploy command won't run if the `build/web` directory isn't created.

Do we need `build:release:output`?

matanlurey: Why? We use a `before_deploy` step.

PR #1003, opened 2018-03-03
Hacker news: build in release mode for deploy

kevmoo:


matanlurey:
Can we figure out how to avoid double deploying?

kevmoo:
@matanlurey – I'll let you run w/ this

PR #1005, opened 2018-03-03
chore(Travis): Add a PWA/deploying stage.

matanlurey:
PTAL @kevmoo, will merge internally if we like it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -314,7 +314,7 @@ jobs:
           - examples/hacker_news_pwa/.dart_tool
       before_deploy:
         - pushd examples/hacker_news_pwa
-        - pub run build_runner build -o build --fail-on-severe
+        - pub run build_runner build -o build --fail-on-severe --config release
         - popd
       deploy:
         skip_cleanup: true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -285,18 +304,23 @@ jobs:
...
+    # See https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Deployments.
+    - stage: deploying
+      script: skip
+      env: PKG="examples/hacker_news_pwa"
+      cache:
+        directories:
+          - examples/hacker_news_pwa/.dart_tool
+      before_deploy:
+        - pushd examples/hacker_news_pwa
+        - pub run build_runner build -o build --fail-on-severe
```
kevmoo: `--config release`

kevmoo: ...actually, now I *really* question the value of doing this...

matanlurey: What do you mean?

kevmoo: Wait. I get it now. Wow. Okay....

PR #1017, opened 2018-03-05
refactor(Examples): A bit of cleanup for HNPWA.

matanlurey:
... using new DI conventions. Also make testable (I'm going to add e2e-tests).

/cc @kevmoo @leonsenft for quick look before internal merge.

### DIFF
--- a/examples/hacker_news_pwa/lib/app_component.dart
+++ b/examples/hacker_news_pwa/lib/app_component.dart
@@ -15,12 +15,13 @@ import 'src/routes.dart';
   directives: const [routerDirectives],
 )
 class AppComponent {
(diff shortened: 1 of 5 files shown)


leonsenft:
I don't like the aesthetics of using static fields in the template itself, but I agree with the choice to make them static.

matanlurey:
@leonsenft:

> I don't like the aesthetics of using static fields in the template itself, but I agree with the choice to make them static.

Ack. @hterkelsen remarked that it would be possible to understand `YourComponent.foo` similar to how we just understand `foo` today. I'll file a FR:

https://github.com/dart-lang/angular/issues/1018

Issue #1020, opened 2018-03-05
Tracking issue: Typed TrackByFn


matanlurey:
This is just a place-holder until I or someone else has time to fill it.

alorenzen:
Dupe of https://github.com/dart-lang/angular/issues/833?

matanlurey:
Duplicate of #833.

chalin:
FYI, the API page for TrackByFn still refers to this issue (which has been closed): https://pub.dartlang.org/documentation/angular/5.0.0-alpha+10/angular/TrackByFn.html

> NOTE: It is not safe to simply assume that the second parameter is of your custom type (neither TrackByFn nor NgFor allow that) at this time: `https://github.com/dart-lang/angular/issues/1020`.

PR #1021, opened 2018-03-05
docs(Core): Add documentation behind 'TrackByFn'.

matanlurey:
(Not for merging, just early feedback)

/cc @chalin @leonsenft @srawlins

### DIFF
--- a/angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
+++ b/angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
@@ -13,9 +13,32 @@ typedef void DefaultIterableCallback(
   int currentIndex,
 );
 
-typedef dynamic TrackByFn(int index, dynamic item);
+/// A function that can be used to return a unique key for [item] at [index].
+///
+/// By default, [item] itself is used as the key to track to instantiate a new
+/// template per item in `*ngFor`. If the data ever changes, AngularDart will
+/// consider any object with a different identity (`identical`) to be different
+/// - and destroy and re-create a new node.
+///
+/// To optimize performance when you have a way to determine uniqueness other
+/// than identity, such as an `id` field on an object returned from the server,
+/// you may specify a [TrackByFn]`:
+/// ```
+/// class MyComp {
+///   Object trackByEmployeeId(int index, Object item) => (item as Employee).id;
+/// }
+///
+/// class Employee {
+///   int id;
+/// }
+/// ```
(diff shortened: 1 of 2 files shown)


angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
```diff
@@ -13,9 +13,32 @@ typedef void DefaultIterableCallback(
...
+/// class MyComp {
+///   Object trackByEmployeeId(int index, Object item) => (item as Employee).id;
+/// }
+///
+/// class Employee {
+///   int id;
+/// }
+/// ```
+///
+/// **NOTE**: It is not safe to simply _assume_ that the second parameter is
```
leonsenft: If this is worth pointing out, should we instead perform an `is` check for type promotion rather than explicitly cast?

chalin: Right. Is something like this what you might have had in mind?

```dart
Object trackByEmployeeId(int _, Object item) => item is Employee ? item.id : item;
```

angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
```diff
@@ -13,9 +13,32 @@ typedef void DefaultIterableCallback(
...
+/// of your custom type (neither [TrackByFn] nor `NgFor` allow that) at this
+/// time: https://github.com/dart-lang/angular/issues/1020.
+typedef Object TrackByFn(int index, Object item);
```
chalin: Just curious: why change the type of `item` from `dynamic` to `Object`? Essentially, the trackBy function is the user suggested way of comparing values for equality, and the `Object.==` is typed as: `bool operator ==(other);`

In terms of #1020, are you thinking of something like this?

```dart
typedef TrackByFn<T,E> = T Function(int index, E item);
```

angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
matanlurey: That's another option, sure.

angular/lib/src/core/change_detection/differs/default_iterable_differ.dart
matanlurey: > Just curious: why change the type of `item` from `dynamic` to `Object`? Essentially

PR #1027, opened 2018-03-06
Form example

alorenzen:
A simple form example using `angular_components` and `angular_forms`.


examples/registration_form/analysis_options.yaml
```diff
@@ -0,0 +1,4 @@
+analyzer:
+  plugins:
+    angular:
+      enabled: true
```
TedSander: Should this have a newline? Here and other files?

examples/registration_form/lib/address/address_component.dart
```diff
@@ -0,0 +1,110 @@
...
+class AddressComponent {
+  String address1;
+  String address2;
+  String city;
+  String state;
+  String zip;
+}
+
+/// A [Validator] which validates that the [requiredFirst] value is preset when
+/// this control has a value set.
```
TedSander: This comment didn't help me too much understanding this validator. Maybe:
Validator that ensures that [requiredFirst] control has a value before this control. For example an address input where given two address lines the second line shouldn't be the only one with a value.

examples/registration_form/lib/address/address_component.html
```diff
@@ -0,0 +1,42 @@
+<div ngControlGroup="address">
+    <material-input
+            type="text"
+            floatingLabel
+            label="Address 1"
+            ngControl="address1"
+            [(ngModel)]="address1"
+            required
+            #address1Control="ngForm">
```
TedSander: Maybe comment this line for users following along? This is a pattern I really like to see, but I think the casual observer would probably not understand what is going on here.

examples/registration_form/lib/address/address_component.html
```diff
@@ -0,0 +1,42 @@
+<div ngControlGroup="address">
+    <material-input
+            type="text"
+            floatingLabel
+            label="Address 1"
+            ngControl="address1"
+            [(ngModel)]="address1"
```
TedSander: I don't think you actually need these ngModels... or at least they don't need to be two way you aren't doing anything with them in the component itself.

examples/registration_form/lib/root/root_component.dart
```diff
@@ -0,0 +1,77 @@
...
+class RootComponent {
+  @ViewChild('myForm')
+  ControlContainer myForm;
+
+  String firstName;
+  String lastName;
+
+  Person person;
+
+  void onSubmit() {
```
TedSander: I think the pattern that I like to see here instead since you aren't really using myForm elsewhere in the component is to pass in the form to this function using a varref instead of using a ViewChild.

examples/registration_form/lib/root/root_component.dart
```diff
@@ -0,0 +1,77 @@
...
+    person = _createPerson();
+  }
+
+  // TODO(alorenzen): Reset form inputs.
+  void onClear() {
+    person = null;
+  }
+
+  Person _createPerson() {
+    return new Person(
```
TedSander: Since you are only using the values I think it is cleaner to do:
Map<String, dynamic> values = form.value;
firstName: values['first-name']

Also some people like having const strings for the values. Instead of using the string everywhere. I personally don't think it is needed, but wanted to point it out if you liked that pattern better.

examples/registration_form/lib/root/root_component.html
```diff
@@ -0,0 +1,57 @@
...
+        </material-input>
+    </section>
+    <address-cmp></address-cmp>
+
+    <material-yes-no-buttons
+            saveCancel
+            yesHighlighted
+            raised
+            [yesDisabled]="!myForm.valid"
+            (yes)="onSubmit()"
```
TedSander: I'd pass in myForm here and the next line.

examples/registration_form/lib/root/root_component.dart
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    providers: [
+      popupBindings,
+    ])
+class RootComponent {
+  @ViewChild('myForm')
+  ControlContainer myForm;
+
+  String firstName;
+  String lastName;
```
TedSander: I'd drop these they aren't needed. Maybe keep one if you want to show ngModel usage, but since you aren't using the values they just add a bit of noise.

examples/registration_form/analysis_options.yaml
alorenzen: Done

examples/registration_form/lib/address/address_component.dart
alorenzen: I definitely struggled a bit with writing this one. Thanks for the suggestion, much better than what I originally came up with.

examples/registration_form/lib/address/address_component.html
alorenzen: Done

examples/registration_form/lib/address/address_component.html
alorenzen: I do eventually want to support a "default" value / editing existing data, so might add these back later. But agree that there's no value now.

examples/registration_form/lib/root/root_component.dart
alorenzen: Done.

examples/registration_form/lib/root/root_component.dart
alorenzen: Ah, I like this much more than what I had originally.

examples/registration_form/lib/root/root_component.html
alorenzen: Don't actually need it for onClear (yet), so left that out for now. I'll be adding it back when I support reset on forms.

examples/registration_form/lib/root/root_component.dart
alorenzen: Done

angular/lib/experimental.dart
```diff
@@ -128,3 +131,17 @@ bool isDomRenderDirty() => app_view.domRootRendererIsDirty;
 void resetDomRenderDirty() {
   app_view.domRootRendererIsDirty = false;
 }
+
+class _ThrowingSlowComponentLoader implements SlowComponentLoader {
+  @override
+  Future<ComponentRef<T>> load<T>(Type type, Injector injector) {
+    throw new UnimplementedError('Don\'t use SlowComponentLoader');
```
TedSander: I don't know if you are going to land this in this CL, but I'd like to see this as a constant since it is used in two places, and also a little more helpful error message.

Probably too much sass but something like this would be cool:
'You are using the faster, slimmer, improved angular. Sadly that means we don't support loading a component with SlowComponentLoader. Please migrate this code to use ComponentFactories or switch back to the X to continue using SlowComponentLoader if you must.'

examples/registration_form/lib/address/address_component.dart
```diff
@@ -0,0 +1,107 @@
...
+    ])
+class AddressComponent {}
+
+/// Validator that ensures that [requiredFirst] control has a value before this
+/// control.
+///
+/// For example an address input where given two address lines the second line
+/// shouldn't be the only one with a value.
+// TODO(alorenzen): Improve the interop between the two Controls. Errors should
+// clear when requiredFirst value is set.
```
TedSander: Just a note this may not be possible currently. I think it might create a circular DI dep since NgControl -> validators and you would need the control to reset the errors.

It might be a really nice feature if either:
a) A validator could have an optional method that gives the control instance when it is set on a control.
b) A validator's interface is expanded to be able to reset itself which gives a callback to the control.

One wrinkle of this is the validator is just one of many so probably all of the validators would need to be rerun on a reset call, or more state would need to be known or assumed to just clear certain errors.

The other way around this is to force validators to not be able to change their states. So the validator could throw if requiredFirst is set twice.

PR #1042, opened 2018-03-08
fix(Compiler): Add better errors when an annotation fails.

matanlurey:


akaufmann:
@matanlurey, @leonsenft, @alorenzen and everyone else involved.

I know this is not the right place, but I must thank you for your hard work. 
It's a pleasure to use AngularDart and look through your PRs/issues every day and see that you are not satisfied with the status quo and working on the right things.
The (new) openness to external users, the great features that are usable and will come, the quick feedback here and on Gitter and your friendliness have led me back to AngularDart. Keep up the good work!

(I know this is adulation but as long as I can't bring more value to your project I have to say/write it 😆)

PR #1052, opened 2018-03-09
Adds documentation on effective styling

leonsenft:
This is a WIP, I'd appreciate anyone's feedback!

### DIFF
--- a/doc/effective/styling.md
+++ b/doc/effective/styling.md
@@ -1,0 +1,230 @@
+# Effective Angular: Styling
+
+AngularDart offers a variety of methods for styling your application. This
+document is intended to help developers understand how and when to use these
+options.
+
+* [Global styles](#global-styles)
+  * [DO declare application-wide styles in
+    index.html](#do-declare-application-wide-styles-in-indexhtml)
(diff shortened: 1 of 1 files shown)


doc/effective/styling.md
```diff
@@ -0,0 +1,230 @@
...
+        font-family: Roboto, sans-serif;
+        font-size: 16px;
+      }
+    </style>
+  </head>
+  <body></body>
+</html>
+```
+
+### DO include external style sheets in index.html
```
matanlurey: Maybe combine this with the previous section, because upon reading it made it seem like application styles should all be `<style>` and external ones should all be `<link>`.

They are sort of interchangable

leonsenft: That message was actually my intention.

I thought I read that declaring styles in a `<script>` tag in the HTML document was preferable to including them in a `<link>` if you could. Something about reducing the number of resources to download. Now I can't seem to find the source, so I skeptical this optimization (if it even is one) is worth mentioning.

TedSander:
I really like it. Looks like just the right amount of advice I would like to see from angular about styling.

Issue #629, opened 2017-09-13
Make AngularDart completely tree-shakeable


AndreyChernykh:
Hello. 

We have discovered a very disappointing behavior of the dart tree shaking in case of the angular-annotated classes.

As you can see, there is no usage of any InjectableService or SimpleService in the code below: 

```dart
/// my_servises.dart
import 'package:angular/di.dart';

@Injectable()
class InjectableService {
  String myProperty = 'Wubalubadubdub';
}

class SimpleService {
  String myProperty = 'Good news, everyone!';
}


/// my_cmp.dart
import 'src/injectable_service.dart';

@Component(
  selector: 'my-app',
  template: '<div></div>'
)
class AppComponent { }
```
If we build that project with `pub build` and look into the compiled `main.dart.js` we will see that the string `'Good news, everyone!'` is not present there, which is definitely good news, because that shows us expected behavior of a tree shaking algorithm.
 
However, the string `'Wubalubadubdub'` was not tree shacked at all. The same behavior is true for any `@Component` class and all its methods or properties; as well as it is true for library files that export bunch of services or components. Like this: 

```dart 
// this_service_is_not_used_by_anyone.dart;
@Injectable()
class InjectableService {}

/// services.dart
export 'this_service_is_not_used_by_anyone.dart';
export 'and_this_service_too.dart';
export 'but_they_will_eventually_be_added_to_the_build_file.dart';

/// my_cmp.dart
import 'src/services.dart';
@Component(...)
class AppComponent { }
```

The situation is even worse if we DO use a class type, but do not use its methods or properties: 
```dart
/// my_cmp.dart
import 'src/services.dart' show InjectableService;

@Component(
  selector: 'my-app',
  template: '<div></div>',
  providers: const [InjectableService]
)
class AppComponent { }
```
In that case, dart analyzer will be unable to track that situation because an `import` is now used.

matanlurey:
@AndreyChernykh:

> We have discovered a very disappointing behavior of the dart tree shaking in case of the angular-annotated classes.

> As you can see, there is no usage of any InjectableService or SimpleService in the code below: 

> ```dart
> /// my_servises.dart
> import 'package:angular/di.dart';
> 
> @Injectable()
> class InjectableService {
>  String myProperty = 'Wubalubadubdub';
> }
>
> class SimpleService {
>  String myProperty = 'Good news, everyone!';
> }
>
>
> /// my_cmp.dart
> import 'src/injectable_service.dart';
>
> @Component(
[...]

PR #1053, opened 2018-03-09
WIP: New runApp/runAppLegacy APIs.

matanlurey:
Closes https://github.com/dart-lang/angular/issues/629.

WIP: Do **not merge**. This doesn't yet pass all test cases, and will have internal breakages as currently implemented. Early review/feedback welcome, otherwise.

Namely, this is a refactor of
* Functions that start your angular application.
* The `PlatformRef` class, which was hard to follow before, and used inconsistently.
* Some experimental functions.


angular/lib/src/bootstrap/run.dart
```diff
@@ -46,6 +41,7 @@ Injector _platformInjector() {
 /// Optionally may supply a [createInjector] function in order to provide
 /// services to the root of the application:
 ///
+/// ```
 /// // Assume this file is "main.dart".
 /// import 'package:angular/angular.dart';
 /// import 'main.template.dart' as ng;
```
chalin: Btw, that's a lot of boilderplate:

```dart
void main() {
  runApp(ng.HelloWorldNgFactory, createInjector: helloInjector);
}

@GenerateInjector(const [
  const ClassProvider(HelloService),
])
final InjectorFactory helloInjector = ng.helloInjector$Injector;
```

Is there any way that this could be simplified to, say:

```dart
@GenerateInjector(const [
  const ClassProvider(HelloService),
])
void main() {
  runApp(ng.HelloWorldNgFactory, createInjector: $Injector);
}
```

angular/lib/src/bootstrap/run.dart
```diff
@@ -85,20 +81,147 @@ ComponentRef<T> runApp<T>(
...
+  });
+}
+
+const _errorMessage = 'Renamed "runAppLegacy". See doc comments for details';
+
+/// See [runAppLegacy].
+///
+/// This method only worked in AngularDart 4.x and below, and relied on the
+/// pub transformer automatically rewriting this function to [bootstrapStatic].
+/// In AngularDart 5.x+, this rewrite should be done manually.
```
chalin: Maybe:

> In AngularDart 5+, this rewrite is done manually.

angular/lib/src/bootstrap/run.dart
```diff
@@ -85,20 +81,147 @@ ComponentRef<T> runApp<T>(
...
+///
+/// @Component(
+///   selector: 'app',
+///   template: 'Hello World',
+/// )
+/// class AppComponent {}
+/// ```
+///
+/// **AFTER**:
+/// ```
```
chalin: There is no `class AppComponent` in the **AFTER** example.

angular/lib/src/bootstrap/run.dart
matanlurey: It could, but then there is more magic.

leonsenft: At the very least you'd still need the `ng` prefix on `$Injector`. But then this would rely on that being some magic top level declaration in the generated file, so what happens if you want multiple generated injectors?

I think the current approach makes it more obvious what the annotation does, and provides you the flexibility to generate multiple injectors with whatever name you desire.

angular/lib/src/bootstrap/run.dart
```diff
@@ -85,20 +81,147 @@ ComponentRef<T> runApp<T>(
...
-  appViewUtils = unsafeCast<AppViewUtils>(appInjector.get(AppViewUtils));
-  sharedStylesHost ??= new DomSharedStylesHost(document);
   final appRef = unsafeCast<ApplicationRef>(appInjector.get(ApplicationRef));
+  appViewUtils = unsafeCast<AppViewUtils>(appInjector.get(AppViewUtils));
   return appRef.bootstrap(componentFactory, appInjector);
 }
+
+/// Starts a new AngularDart application in legacy mode, with [initReflector].
+///
+/// This method was formally known as `bootstrap` or `bootstrapStatic`, and
```
4cm4k1: formally => formerly ;)

matanlurey: Thanks!

matanlurey:
Just an update, I'm helping with some Dart2 migrations, and I'll be back to complete this work, probably second-half of the week. So this is on hold for now.

PR #1056, opened 2018-03-10
feat(Runtime): Give a better error on Output/ cast failures.

matanlurey:


angular/lib/src/core/linker/app_view.dart
```diff
@@ -654,10 +661,31 @@ abstract class AppView<T> {
...
       markPathToRootAsCheckOnce();
-      appViewUtils.eventManager.getZone().runGuarded(() => handler(event as F));
+      _ngZone.runGuarded(() {
+        handler(event as F);
+      });
+    };
+  }
+
+  /// Variant of [eventHandler1] that provides better runtime errors.
+  void Function(E) _eventHandler1Debug<E, F extends E>(void Function(F) fn) {
```
leonsenft: So we could just put this in `DebugAppView.eventHandler1()` instead. Is this future-proofing for when we remove all the debug-related code?

matanlurey: Yes. DebugAppView already does not have event handlers.

angular/lib/src/runtime/errors.dart
```diff
@@ -0,0 +1,18 @@
...
+class InvalidHandlerError extends TypeError {
+  final Type expected;
+  final Type actual;
+  final Object context;
+
+  InvalidHandlerError(this.expected, this.actual, this.context);
+
+  @override
+  String toString() =>
+      'Invalid event handler on $context: Expected an event of type $expected '
```
leonsenft: Saying the event handler is invalid isn't necessarily correct. Many of the errors I fixed were caused by emitting the wrong types from the stream the handler was registered to.

For example, a handler of type `(List<Foo>) => void` was registered to a stream of type `Stream<List<Foo>>`, but the client code was actually adding values of type `List` to the stream.

I've forgotten, will this sort of invalid implicit cast become an error as well? If so then I guess this message is correct, as it's the only way we'll get here.

matanlurey: > Saying the event handler is invalid isn't necessarily correct. Many of the errors I fixed were caused by emitting the wrong types from the stream the handler was registered to.

angular/lib/src/core/linker/app_view.dart
leonsenft: Oh yeah it's not as simple as wrapping the call to `super`, since the check has to happen inside the returned closure, so you'd need to fully rewrite the code in the debug view anyways, which would probably be more prone to error and getting out of sync. This approach is better.

angular/lib/src/runtime/errors.dart
leonsenft: Yeah thinking about this more it seems like this _could_ provide a misleading error _until_ invalid implicit casts become a run-time error, but after that it's correct.

Are you fine writing it for the future now? Or do you want an intermediate message?

PR #1082, opened 2018-03-16
WIP: Use AstDirectiveNormalizer. Not all tests pass.

matanlurey:
Here is the latest result (locally) of running the tests:
https://gist.github.com/matanlurey/8abcf3d8f2a3dc01ef6a2f37e56066d3

It looks like the expression parsing, specifically, is not the same.


matanlurey:
I see one crash too. This causes a crash, when parsed:
```html
  <child [value]="\'a\n\nb\'"></child>
```

I think it is either the escaped `'` or the new-line characters.

Issue #963, opened 2018-02-27
Improve HN app perceptual speed index


filiph:
A big part of PWA scoring is something called "perceptual speed index". This is best explained [here](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/metrics/speed-index). Basically, the earlier we can show the UI (even if it's only a placeholder) the better. 

Ideas:

* Show the top bar in the static HTML
* Add static placeholders for the loaded articles with white divs

The static page (before even starting to download JS) should look like this:

![slice](https://user-images.githubusercontent.com/919717/36761085-5ceb5bfc-1bd2-11e8-9303-a26ef9f80d58.png)

Or at least like this:

![slice2](https://user-images.githubusercontent.com/919717/36761150-a402298a-1bd2-11e8-9f1f-4f612d683968.png)

In general, the earlier we can have at least some of the pixels correspond to the final render, the better.

This is not just about "beating a score", by the way. The apps that do this kind of optimization _do_ perceptually load faster. It's a very different experience looking at a gray screen for 2 seconds than looking at (unfilled) beginnings of a UI.

PR #1084, opened 2018-03-18
feat(Examples): Increase the perceptual-speed-index of HNPWA.

matanlurey:
A number of changes were made here. Let me explain them:

* Moved the common application styles into index.html, disabling encapsulation.
  * This loads more styles before the application loads.
  * This allows a sort of SSR-lite, showing the shell of the app before JS loads.
* Prefetches and in-memory-caches the current URL/RPC before the app totally loads.
  * This starts requesting the data from server as soon as `main()` is hit.
  * This avoids any [FOUC][], and makes the application appear to "pop in".
* Replaces pkg/http with plain use of dart:html to drop code-size.

There are probably a bit better ways to do all of this, but it's a start, and was pretty low effort. Other ideas include prefetching the RPC in a 1-line pre-amble JavaScript (even before `main()` is hit), but again, lower impact than this.

[FOUC]: https://en.wikipedia.org/wiki/Flash_of_unstyled_content

Closes https://github.com/dart-lang/angular/issues/963.


kevmoo:
Perhaps we can merge our efforts?

https://github.com/dart-lang/angular/pull/1085

examples/hacker_news_pwa/web/index.html
```diff
@@ -5,6 +5,7 @@
   <meta name="theme-color" content="#212121">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta name="Description" content="A Hacker News reader implemented as a Progressive Web App using AngularDart.">
+  <link rel="prefetch" href="https://node-hnapi.herokuapp.com/news?page=1" as="fetch" crossorigin />
```
kevmoo: Why "prefetch" instead of "preload?

matanlurey: Preload isn't appropriate unless its 100% going to hit that URL.

(Chrome Dev Console warns you as such)

PR #1085, opened 2018-03-18
Remove use of pkg:http, preload first news page

kevmoo:
before: 405,705, after: 336,164
69K, 17% smaller - uncompressed

2s faster to interactive: https://www.webpagetest.org/video/compare.php?tests=180318_QQ_72ed89af4587e6a803d3d39338640d7c,180317_ZY_d2c37ca1631c7a550d1e73d3f71ff518

### DIFF
--- a/examples/hacker_news_pwa/lib/hacker_news_service.dart
+++ b/examples/hacker_news_pwa/lib/hacker_news_service.dart
@@ -1,8 +1,8 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:html';
 
 import 'package:angular/angular.dart';
-import 'package:http/http.dart';
 
 /// Represents the base URL for HTTP requests using [HackerNewsService].
 const baseUrl = const OpaqueToken<String>('baseUrl');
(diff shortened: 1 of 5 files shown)


examples/hacker_news_pwa/lib/hacker_news_service.dart
```diff
@@ -1,28 +1,27 @@
...
 /// Represents the base URL for HTTP requests using [HackerNewsService].
 const baseUrl = const OpaqueToken<String>('baseUrl');
 const defaultBaseUrl = 'https://node-hnapi.herokuapp.com';
 
 class HackerNewsService {
   final String _baseUrl;
-  final BaseClient _client;
 
-  HackerNewsService(@baseUrl this._baseUrl, this._client);
+  HackerNewsService(@baseUrl this._baseUrl);
```
matanlurey: We need to be able to write unit tests for this code, so the HTTP service needs to be injectable or otherwise overridable for testing. Unless we've decided pkg/http shouldn't be used anymore because of code bloat I'd rather continue to push on getting the size down there.

examples/hacker_news_pwa/web/index.html
```diff
@@ -9,6 +9,7 @@
   <title>Hacker News - AngularDart PWA</title>
   <script defer src="main.dart.js"></script>
   <link rel="manifest" href="manifest.json">
+  <link rel="preload" href="https://node-hnapi.herokuapp.com/news?page=1" as=fetch crossorigin >
```
matanlurey: This is interesting. I'm concerned a bit its a "waste" if not navigating to that page, but I think for a demo (which this is) it's probably the right call. I'll move this into my PR.

examples/hacker_news_pwa/lib/hacker_news_service.dart
kevmoo: We can always create a trivial, injectable class.

I can't imagine a way to make `pkg:http` smarter in the trivial case.

There is **so** much plumbing between newing up a client and hitting send – to expect dart2js to figure out that at runtime we only need trivial `get` – no headers, no peeking at response status code – I think it's unrealistic.

matanlurey: Fair enough. I merged this change into my PR (#1084).

matanlurey:
Closing in favor of #1084 which has most of this copied over.

PR #1092, opened 2018-03-20
Run VM tests separately

natebosch:
In an upcoming release of package:test it will refuse to run VM tests
with the `--precompiled` flag to prevent usage mistakes.

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -49,14 +49,18 @@ case $TASK in
 
   test)
     echo -e '\033[1mTASK: Testing [test]\033[22m'
-    echo -e 'pub run build_runner test --fail-on-severe -- -P travis'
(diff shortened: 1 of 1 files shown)


matanlurey:
As I suspected :( @natebosch

natebosch:
I'm confused how these tests were passing before, the `--precompiled` argument was previously ignored on the VM platform...

matanlurey:
Right, but I think (some) of them relied on being able to read generated outputs.

natebosch:
Ah, I think some of the failures are because when you run `pub run test -p vm` it ignores `dart_test.yaml` saying there are only web tests...

natebosch:
package:test will warn instead of fail. Will need to revisit a long term solution here but we don't need this for now.

PR #1093, opened 2018-03-20
WIP: Use ChangeDetectionHost for ApplicationRef.

matanlurey:
I deleted `application_ref.test`, mostly because it didn't do much, and fails now.

I plan to write a new set of tests for `ApplicationRef` before I land this PR.


matanlurey:
This passed internal testing (down to ~1 failure after some fixes). I'll add tests now.

Issue #553, opened 2017-07-24
Should we infer @Optional based on argument type?


matanlurey:
Had a nice discussion today, why do we even _have_ `@Optional`?

```dart
class Animal {
  Animal(@Optional() Ecosystem ecosystem);
}
```

... since Dart has this already, in both positional and named arguments:

```dart
class Animal {
  Animal([Ecosystem ecosystem]);
}
```

```dart
class Animal {
  Animal({Ecosystem ecosystem});
}
```

It seems safe to say we could _infer_ all optional parameters as `@Optional()`, and remove this key word entirely. Two potential wrinkles though that need further discussion:

1. How do we treat named arguments annotated with `@required`?
2. How do we treat named or positional optional arguments with default values?

i.e.:

```dart
class Animal {
  Animal({@required Ecosystem ecosystem});
}
```

My intuition here is to treat this as required (not optional).

```dart
class Animal {
  Animal({Ecosystem ecosystem: const DefaultEcosystem()};
}
```

My intuition here is to either:

a. Ignore and do not support for DI
b. Try to use the same "reviving" logic we use for `useValue`:

```dart
new Animal(ecosystem: injector.getOptional(Ecosystem) ?? const DefaultEcosystem())
```

A final question is whether we need something like `@ignore`. My intuition is no.

matanlurey:
I think the optional named/unnamed parameters change is likely a much bigger change, and not in scope for 5.x unless we are very motivated (it would require re-ordering parameters in client code). I think that rolling it our incrementally as an optional feature for ^5.1.0 is more likely. Ideas:

```dart
class A {
  @Inject.useOptionalParameters()
  A(B b, [C c]);
}
```

... other names, also welcome :)

leonsenft:
I'd be okay with simply ignoring `@required` or perhaps warning users it won't be honored.

I see the annotation's purpose as a way of using named parameters to make your functions self-documenting at their call-site. But with Angular, users are almost never exposed to the call-sites of component constructors, so it loses its purpose.

I'm in favor of using the same reviving logic to use the default values when injection fails.

leonsenft:
We could even choose to only support unnamed optional parameters to avoid the issue of `@required`, based on the same argument that named parameters are self-documenting at call-sites, which again don't matter in our case.

matanlurey:
@leonsenft Luckily we don't support named arguments at all (https://github.com/dart-lang/angular/issues/448), so we can skip that for now.

[...]

PR #1186, opened 2018-04-05
feat(DI): add decorators without parenthesis

luisvt:
Add shorthand decorators that do not have parenthesis. This way users would not need to use the parenthesis whenever the decorator are used. For example, instead of:

```dart
@Injectable()
class SomeService {
...
}```

users could write:

```dart
@injectable
class SomeService {
...
}```

### DIFF
--- a/angular/lib/src/core/di/decorators.dart
+++ b/angular/lib/src/core/di/decorators.dart
@@ -104,6 +104,9 @@ class Injectable {
   const Injectable();
 }
 
+/// Shorthand of [Injectable].
+const Injectable injectable = const Injectable();
+
 /// A parameter metadata that marks a dependency as optional.
 ///
 /// [Injector] provides `null` if the dependency is not found.
@@ -125,6 +128,9 @@ class Injectable {
 class Optional {
   const Optional();
 }
+
+/// Shorthand for [Optional].
+const Optional optional = const Optional();
 
 /// Specifies that an [Injector] should retrieve a dependency only from itself.
 ///
@@ -152,6 +158,9 @@ class Self {
   const Self();
 }
 
+/// Shorthand for [Self].
+const Self self = const Self();
+
 /// Specifies that the dependency resolution should start from the parent
 /// injector.
 ///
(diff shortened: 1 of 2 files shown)


zoechi:
What about `@Host()` https://github.com/luisvt/angular/blob/fa0498555d251623295d38bcc6844b6a3a334fb8/angular/lib/src/core/di/decorators.dart#L246 ?
I guess the first line of doc comments should start with uppercase and  end with a `.` (not 100% sure what the style guides are here)

luisvt:
most likely I forgot it, let me add it

matanlurey:
I'll have to check that this isn't too difficult to land internally (it could conflict with local references or symbols). In general though I'm supportive, though:

* `@Injectable()` will be soft-deprecated in 5.x (no longer needed for most customers)
* `@Optional()` we're hoping to replace with Dart semantics (#553).
* ... for `@skipSelf`, `@self`, etc, I'm actually considering merging them into a single object:
  * `@Inject.fromSelfOnly()`
  * `@Inject.skipSelf()`
  * `@Inject.fromParentOnly()`

... to make it more clear what those annotations do. WDUT?

luisvt:
@zoechi I added the `@host` annotation.

@matanlurey it would be good to have shorthands for those annotations, something like:

* `@Inject.fromSelfOnly()`  -->  `@injectFromSelfOnly` or `@inject_fromSelfOnly`
* `@Inject.skipSelf()`  -->  `@injectSkipSelf` or `@inject_skipSelf`
* `@Inject.fromParentOnly()`  -->  `@injectFromParentOnly` or `@inject_fromParentOnly`

jolleekin:
Those annotations work in the context of DI, so it seems redundant to prefix them with `inject` (`@host` is an example). I would propose the following
- `@fromSelf`
- `@skipSelf`
- `@fromParent`

By the way, are `@skipSelf` and `@fromParent` the same?

matanlurey:
We've decided we aren't moving forward with this at this time.

We've already ran over our migration budget for v5, we can revisit after v5 is released.

luisvt:
😞 I'll wait for next version then

Issue #1132, opened 2018-03-25
Inconsistent naming conventions for enum values


jolleekin:
According to [this style guide][1], constant variables, including enum values, should be in *lowerCamelCase*. However, there are still many enums that don't follow this guide. Some examples are:

``` dart
enum ViewType {
  HOST,
  COMPONENT,
  EMBEDDED
}

enum ViewEncapsulation {
  Emulated,
  None
}

enum LifecycleHooks {
  OnInit,
  OnDestroy,
  DoCheck,
  OnChanges,
  AfterChanges,
  AfterContentInit,
  AfterContentChecked,
  AfterViewInit,
  AfterViewChecked
}

enum HtmlTokenType {
  TAG_OPEN_START,
  TAG_OPEN_END,
  TAG_OPEN_END_VOID,
  TAG_CLOSE,
  TEXT,
  ESCAPABLE_RAW_TEXT,
  RAW_TEXT,
  COMMENT_START,
  COMMENT_END,
  CDATA_START,
  CDATA_END,
  ATTR_NAME,
  ATTR_VALUE,
  DOC_TYPE,
  EXPANSION_FORM_START,
  EXPANSION_CASE_VALUE,
  EXPANSION_CASE_EXP_START,
  EXPANSION_CASE_EXP_END,
  EXPANSION_FORM_END,
  EOF
}

enum _PreparsedElementType { NG_CONTENT, STYLE, STYLESHEET, SCRIPT, OTHER }

enum NavigationResult { SUCCESS, BLOCKED_BY_GUARD, INVALID_ROUTE }
```

AngularDart 5.0.0-alpha+8

[1]:  https://www.dartlang.org/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

matanlurey:
Thanks for filing.

Some of these are in the internal API only (i.e. not importable by the user), so migrating them is both easier but also lower priority. Pull requests welcome for these:

* `ViewType`
* `LifecycleHooks`
* `HtmlTokenType`
* `_PreparsedElementType`

For the ones that _are_ importable by the user, we will require migrations internally, which are probably not worth it for this release, though we can track them here:

* `ViewEncapsulation`
* `NavigationResult`
* `ChangeDetectorState`
* `ChangeDetectionStrategy`

matanlurey:
I think this is unlikely to happen given the effort and what the pay-off is (virtually none).

PR #1209, opened 2018-04-07
refactor(StyleGuide): follow style guide for private enum field

Kleak:
rename private enum field to follow the style guide

this is a part of the #1132 issues


matanlurey:
Thank you for the contribution! I'll validate that this doesn't break anything unexpected.

PR #1218, opened 2018-04-10
chore(Examples): use --release to build with dart2js

### LINKED ISSUE #1213: Remove all of the build.release.yaml, use --release
1) Set min dependency on build_runner to 0.8.2 and build_web_compilers to 0.3.6
2) Configure dart2js in build.yaml iff you want special config https://github.com/dart-lang/build/blob/master/docs/getting_started.md#switching-to-dart2js
   otherwise, do onthing
3) Set up any auto-build to pass in `--release` instead of `--config release`
4) Delete `build.release.yaml`

Kleak:
should close #1213


matanlurey:
Please rebase on master.

examples/hello_world/build.debug.yaml
```diff
@@ -1,12 +0,0 @@
-targets:
```
jakemac53: I believe you can still get this with a `debug_options` section in the build.yaml, just fyi

cc @natebosch for confirmation (I think we still need docs for this)

natebosch: You can use something like:

```yaml
targets:
  $default:
    builders:
      build_web_compilers|entrypoint:
        generate_for:
           - web/main.dart
        options:
          compiler: dart2js
          dart2js_args:
            - --fast-startup
            - --trust-type-annotations
            - --trust-primitives
        release_options:
          dart2js_args: # inherits `compiler` from default options, and overrides the args
            - --fast-startup
            - --minify
            - --trust-type-annotations
            - --trust-primitives
```

And if you want an option to only apply in non-release (null value in release) you can use the key `dev_options` which also overrides the defaults from `options` only in non-release builds

matanlurey:
@natebosch LGTM?

Issue #974, opened 2018-02-28
Update Previous & Next buttons to be "pretty"


kevmoo:
Example: <img width="196" alt="screen shot 2018-02-28 at 12 00 46 am" src="https://user-images.githubusercontent.com/17034/36776416-89adebc6-1c1a-11e8-828c-cb46dbf1675e.png">

We just have a more button now – not styled. We can do better

leonsenft:
On this topic, the API we're using doesn't give us a count of how many items them are in a particular feed, so we can't in all cases ever disable the next link / show the max number of pages.

Switching to use the official API would help, although I don't think it's suitable for direct consumption in the browser, since you have to first make a request for all the article IDs in a feed, then fetch their data.

What most implementations have done are use the Firebase real-time database to keep the data in memory on a server, which can then be served it to the client in the most convenient format possible on request.

https://github.com/tastejs/hacker-news-pwas#data-sources

PR #1250, opened 2018-04-14
feat(Example): fixing previous/next buttons

kevmoo:
Closes https://github.com/dart-lang/angular/issues/974

### DIFF
--- a/examples/hacker_news_pwa/lib/src/feed_component.html
+++ b/examples/hacker_news_pwa/lib/src/feed_component.html
@@ -18,11 +18,11 @@
 
 <footer>
   <nav>
-  <div *ngIf="previousPageUrl != null"  >
-    <a [routerLink]="previousPageUrl">Previous</a>
-  </div>
-  <div class="next" *ngIf="nextPageUrl != null">
-    <a [routerLink]="nextPageUrl">Next</a>
-  </div>
+    <div *ngIf="previousPageUrl != null"  >
+      <a [routerLink]="previousPageUrl">Previous</a>
+    </div>
+    <div class="next" *ngIf="nextPageUrl != null">
+      <a [routerLink]="nextPageUrl">Next</a>
+    </div>
   </nav>
 </footer>
--- a/examples/hacker_news_pwa/lib/src/feed_component.html
+++ b/examples/hacker_news_pwa/lib/src/feed_component.html
@@ -1,6 +1,6 @@
 <!-- Loading view -->
 <template [ngIf]="items == null">
-  <div>Loading...</div>
+  <div class="loading">Loading...</div>
 </template>
 
 <!-- Loaded empty view -->
@@ -14,6 +14,8 @@
   <span class="rank">{{i + startingRank}}</span>
   <item [item]="item"></item>
 </article>
+</template>
+
 <footer>
   <nav>
   <div *ngIf="previousPageUrl != null"  >
@@ -24,4 +26,3 @@
   </div>
   </nav>
 </footer>
-</template>
--- a/examples/hacker_news_pwa/lib/src/feed_component.html
+++ b/examples/hacker_news_pwa/lib/src/feed_component.html
@@ -14,7 +14,14 @@
   <span class="rank">{{i + startingRank}}</span>
   <item [item]="item"></item>
 </article>
-<div>
-  <a [routerLink]="nextPageUrl">Next page</a>
-</div>
+<footer>
+  <nav>
+  <div *ngIf="previousPageUrl != null"  >
+    <a [routerLink]="previousPageUrl">Previous</a>
+  </div>
+  <div class="next" *ngIf="nextPageUrl != null">
+    <a [routerLink]="nextPageUrl">Next</a>
+  </div>
+  </nav>
+</footer>
 </template>
--- a/examples/hacker_news_pwa/lib/src/feed_component.html
+++ b/examples/hacker_news_pwa/lib/src/feed_component.html
@@ -1,11 +1,11 @@
 <!-- Loading view -->
 <template [ngIf]="items == null">
-  <div class="loading">Loading...</div>
+  <div class="notice">Loading...</div>
 </template>
 
 <!-- Loaded empty view -->
 <template [ngIf]="items != null && items.isEmpty">
-  <div>There are no more items.</div>
+  <div class="notice">There are no more items.</div>
 </template>
 
 <!-- Loaded items view -->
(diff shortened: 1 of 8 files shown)


kevmoo:
screen shot 2018-04-13 at 8 52 45 pm

kevmoo:
100*(1-343261/354496) = 3.2% size increase 😕

kevmoo:
Live sample here https://gh-pages--affectionate-austin-b56d01.netlify.com/

matanlurey:
According to @leonsenft this isn't functioning properly, right?

examples/hacker_news_pwa/web/index.html
```diff
@@ -15,17 +15,25 @@
...
       font-family: Roboto, sans-serif;
       font-size: 18px;
+      height: 100%;
+      margin: 0;
+      display: flex;
+      flex-direction: column;
     }
     body {
-      background: #f2f2f2;
+      background: #212121;
```
leonsenft: What is this even styling? It doesn't seem to make a difference.

kevmoo: When you overscroll on a Desktop machine, the body background is used to paint the overscroll region.

This makes the overscroll region match the header/footer

examples/hacker_news_pwa/lib/src/feed_component.html
```diff
@@ -1,6 +1,6 @@
 <!-- Loading view -->
 <template [ngIf]="items == null">
-  <div>Loading...</div>
+  <div class="loading">Loading...</div>
```
leonsenft: If you're going to try and style up these loading state let's do the same for zero state. Check out the Jobs feed and click next until you get an empty page. Perhaps give this class a less-loading specific name and apply it to the div below and share the styles?

examples/hacker_news_pwa/web/index.html
```diff
@@ -15,17 +15,25 @@
...
       margin: 0;
       overflow-wrap: break-word;
       -webkit-font-smoothing: antialiased;
       overflow-y:scroll;
     }
+    app {
+      flex: 1;
```
leonsenft: Again what does flex here even do?

examples/hacker_news_pwa/web/index.html
```diff
@@ -15,17 +15,25 @@
...
+      background: #eee;
```
leonsenft: This doesn't seem necessary if you undo your background color change which I don't quite follow.

examples/hacker_news_pwa/web/index.html
leonsenft: But now the body also gets painted that color when you over scroll horizontally. 🤕 

How about this instead?
```css
body {
  overscroll-behavior: none;
}
```

leonsenft: Actually that doesn't feel great. Honestly my preference is to leave this unchanged.

leonsenft: Aha solution!

Keep your new styling, plus add to the `body` styles

```css
 overscroll-behavior-x: none;
```

examples/hacker_news_pwa/lib/src/feed_component.html
```diff
@@ -14,7 +14,15 @@
...
   <item [item]="item"></item>
 </article>
-<div>
-  <a [routerLink]="nextPageUrl">Next page</a>
-</div>
 </template>
+
+<footer>
+  <nav>
+  <div *ngIf="previousPageUrl != null"  >
```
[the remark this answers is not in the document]
kevmoo: Can't. The divs split the view left/right. They are styled as `flex`. You don't want the `a` tags as flex because then their clickable size extends beyond the text.

And in the case where there is only one next/previous the clickable area in the whole width of the footer...

examples/hacker_news_pwa/web/index.html
```diff
@@ -64,6 +72,10 @@
       margin: 56px auto 0 auto;
       max-width: 800px;
     }
+    feed {
```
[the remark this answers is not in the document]
kevmoo: Just removed.

examples/hacker_news_pwa/web/index.html
kevmoo: I just did this – saw no delta in the behavior on my laptop.

Does it do anything on mobile?

kevmoo: Actually, it does. ` overscroll-behavior-x: none;` prevents me from swiping left/right to navigate!

kevmoo:
Redeployed here https://gh-pages--affectionate-austin-b56d01.netlify.com/

@leonsenft PTAL

examples/hacker_news_pwa/lib/src/feed_component.html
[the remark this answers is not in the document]
leonsenft: Fair enough!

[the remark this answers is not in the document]
leonsenft: nit: if you're keeping them, mind indenting them please?

examples/hacker_news_pwa/web/index.html
leonsenft: Ugh, it prevents over scrolling horizontally which then showed the dark body color on the sides and looked wonky. I didn't realized this prevented navigation 😭

examples/hacker_news_pwa/lib/src/feed_component.html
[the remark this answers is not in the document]
kevmoo: done

leonsenft:
Looking great, I like the removal of the grey actually.

One note on not duplicating that style you removed in https://github.com/dart-lang/angular/pull/1250/commits/ff22ffb77efa0e62da14b4e75cc22528108f48a9. It's sort of accidental that the feed styles are working there. You can't actually rely on the feed styles for the item detail, since the feed isn't on screen with the item detail at the same time.

When Angular loads a component for the first time, it loads its styles into the DOM, and they stay there even after it's destroyed.

Since you're probably loading the app at the index page, you have to click through a feed to reach an item detail page. That means the feed styles are still present. However, if a user navigates directly to an item page (perhaps linked externally), the feed will have never loaded, thus those styles won't apply to the item detail.

The fix is to move those styles from feed_component.css up to app_component.css.

kevmoo:
> The fix is to move those styles from feed_component.css up to app_component.css.

done!

leonsenft:
Alright, last nit. Mind adding the footer to the static HTML in index.html and moving its styles there as well?

If you do a hard refresh of the index page (or visit the first time), you'll notice all of the shell is there _except_ the footer which snaps in afterwards.

kevmoo:
Done!

leonsenft:
Awesome, thanks a lot for the improvements!

Issue #1282, opened 2018-04-25
Add a PostProcessBuilder which removes placeholder files and .css and .html


natebosch:
My initial thought is we should:

- Remove all `*.ng_placeholder` files, always
- In release (deployed) mode by default, remove `*.css` and `*.html` files from the `lib/` directory
- Allow configuring a list of globs to keep `*.css` or `*.html` files that are needed at deploy time (but are unused by AngularDart). For example, if you rely on `lib/global.css` instead of `web/global.css`.

matanlurey:
To clarify, this is for release builds (where we would be deploying the artifacts).

I'd love the input of some of the heavier AngularDart users externally, because from my perspective this totally makes sense, but maybe we are missing something or some defaults that would help most users.

jakemac53:
Note that you could use a heuristic to only delete the files that seem obviously safe to remove.

- For `.html` files you could check if they look like an angular template, or at least don't look like an entrypoint (don't have a `<html>`, `<head>`, `<body>`, or `<script>`) for instance.
- For `.css`  files you could check for a corresponding `.css.shim.dart` file, and/or do as suggested above and only remove files under `lib`.

natebosch:
> you could check for a corresponding 

Not with the current `PostProcessBuildStep` interface. We only give access to the primary input.

corporateselect:
we already have a quite big angular project now and we rely on some css in lib that is unused by ang - but instead of configuring a list for that globs isnt a more convention over configuration approach better? like a directory, prefix, ... ? Or could there be disadvantages im missing

matanlurey:
One more question. For _release_ builds, is there _any_ reason to retain the `lib/` folder of any package at all? Or am I missing something?

corporateselect:
>Does that sound OK?

sure, that sounds perfect. was thinking bit selfish :)

>One more question. For release builds, is there any reason to retain the lib/ folder of any package at all? Or am I missing something?

Was thinking about that too, like i said we have html/css in lib that we want to retain right now - but, were thinking about to separate it to web/ because it seems a cleaner approach

natebosch:
> is there _any_ reason to retain the `lib/` folder of any package at all?

[...]

PR #1287, opened 2018-04-27
Use -o web:build for deployment

kevmoo:
Eliminate symlinks, creates a cleaner output directory


kevmoo:
Going to hold off on this until https://github.com/dart-lang/angular/issues/1282 is fixed

PR #1317, opened 2018-05-15
Support the latest pkg:analyzer

kevmoo:


alorenzen:
What version is landed internally?

kevmoo:
> What version is landed internally?

Even newer than this...

Issue #1354, opened 2018-05-30
Error during change detection while routing crashes app


matanlurey:
It's possible, likely due to (a) the semantics of the router and (b) the manual use of `.detectChanges()`, to get into an infinite exception loop that will crash/hang the browser. Basically, if a routed-to component has a getter, in its template, that throws, the component will never be set to `ChangeDetectorState.Errored` _and_ it will continuously be created over and over.

Here is the starting point for a test case I am writing to catch and iterate on a solution:

```dart
@TestOn('browser')
import 'package:angular/angular.dart';
import 'package:angular_router/angular_router.dart';
import 'package:angular_router/testing.dart';
import 'package:test/test.dart';

import '1354_infinite_exceptions_test.template.dart' as ng;

void main() {
  test('should ...', () {
    runApp(ng.AppComponentNgFactory);
  });
}

@Component(
  selector: 'app',
  template: r'''
    Sample Application

    <router-outlet [routes]="routes"></router-outlet>
  ''',
  directives: const [
    RouterOutlet,
  ],
  providers: const [
    routerProvidersTest,
  ],
)
class AppComponent {
  static final routes = [
    new RouteDefinition(
      path: 'view',
      component: ng.ViewComponentNgFactory,
    ),
  ];

  final MockLocationStrategy locationStrategy;

  AppComponent(@Inject(LocationStrategy) this.locationStrategy);
}

@Component(
  selector: 'view',
  template: 'Hello World',
)
class ViewComponent {}
```

matanlurey:
OK, so far I haven't been able to re-produce exactly, when I route to this:

```dart
@Component(
  selector: 'throws',
  directives: const [
    NgIf,
  ],
  template: r'''
    <ng-container *ngIf="someCondition">
      Should not be shown.
    </ng-container>
  ''',
)
class ThrowingComponent {
  bool get someCondition {
    throw new IntentionalException();
  }
}

class IntentionalException implements Exception {}
```

... I assume this is because I'm missing some other corner cases.

In the reported user stack trace, they had the following (in Dart2JS):

```
    at Object.wrapException
    at Service._somethingThatThrows$2
    at Service.get$listOfEntities
    at _ViewComponent1.detectChangesInternal$0
    at RouterOutlet_prepare_closure.call$0
    at JsLinkedHashMap.putIfAbsent$2
    at RouterOutlet.prepare$1
```

... which is different from what I'm seeing (mind you, in DDC):

```
[...]

PR #1357, opened 2018-05-31
fix(Router): Wrap usages of `.detectChanges()` to avoid a crash.

matanlurey:
Closes https://github.com/dart-lang/angular/issues/1354.


matanlurey:
This was replaced by @leonsenft

Issue #1267, opened 2018-04-23
Test & update/remove the examples in the API docs


kwalrath:
I happened to notice that a code snippet in an API doc comment (https://github.com/dart-lang/angular/blob/master/angular/lib/src/core/di/decorators.dart#L175) uses `throws`, which has been deprecated in favor of `throwsA()`. If we tested and auto-included the examples in dartdocs, this would've been caught.

/cc @chalin

matanlurey:
No longer planned.

Issue #996, opened 2018-03-02
Validate that code snippets in docs are at least "analyzer clean"


matanlurey:
I believe (?) @chalin and others had a tool for this at one point.

Basically I'd have love to catch that:

```md
const ClassProvider(Foo, ConcreteFoo);
```

... was an incorrect example (missing the `useClass: ` named parameter).

Might not be worth doing, though.

chalin:
If this is related to #992, I was going to suggest that we integrate that page into [site-webdev](https://github.com/dart-lang/site-webdev) since we have there the necessary infrastructure allowing code excerpts to be analyzed and/or tested.

Actually, as @kwalrath mentioned already, it might be a good idea to have an **Effective Angular** section under site-webdev, similar to the **[Effective Dart][]** section under dartlang.org.

[Effective Dart]: https://www.dartlang.org/guides/language/effective-dart

FYI, regarding tooling, the Flutter site has some tooling to extract code excerpts and then analyze them. Under site-webdev, we use the usual tools to analyzer and test code. Instead we have a code excerpt extractor that scans the code, picks out fragments, and embeds the fragments into the docs. The approaches are complementary.

matanlurey:
As mentioned elsewhere, I'd really like this to be in our own repository. This stuff changes often, and the cognitive load of needing to land the CL separate to updating the best practices doesn't seem worth it.

alorenzen:
Also, if it is in our own repository, we can reference it easier internally as well.

Issue #942, opened 2018-02-27
Should docs and examples for webdev.dartlang live here?


matanlurey:
... instead of in another repository we don't really review often?

/cc @chalin @kwalrath @kevmoo

chalin:
Good question. Some initial thoughts below.

Are there specific issues that we're trying to address by this? E.g., might there be a feeling that the docs and/or examples aren't keeping up closely enough with the rapid pace of Angular changes?

> Aside: Along similar lines, why not bring `angular_components` under the `dart-lang/angular` umbrella too --- I'd love to see ACX releases in sync with NG releases. /cc @nshahan 

That reminds me, webdev also hosts some (minimal) ACX docs, as well as other non-angular Dart-for-web examples and docs. We also host a full suite of API docs, though it seems that API docs will (so?) be hosted via the pub site, so that might change.

As you know there are pros and cons. Some of the **pros** of having a **separate repo** (off the top of my head):

- webdev is **off the critical path**, meaning **Angular releases won't be delayed** by issues with the examples and/or docs.
- webdev currently has 30+ examples, some of which use ACX; these examples are built and (component and E2E) tested under ddc and dart2js. E2E currently uses [Protractor](http://www.protractortest.org). As a separate repo, doing all this doesn't slow down `angular` builds -- and this seemed to be a concern recently.
- The repo is currently fully external, hosted on GH, so there are no issues of having to merge PRs from GH internally, and then pushing updates back out to GH.
- Repo access permissions can easily different for webdev. 

Of course, it doesn't have to be all-or-nothing: maybe _some_ core examples could be under `dart-lang/angular`. Or maybe it could be all of the examples, but leave the docs separate.

AngularTS have docs and examples under the `angular/angular` repo (but then again, they serve the docs using an Angular app, so that's extra motivation to ensure that that app is always fully operational with each AngularTS release).

The "pros" listed above become "cons" once the docs and examples are put under `dart-lang/angular`, but if those cons don't bother the core Angular team, then it might indeed be worth considering.

nshahan:
It would be interesting if we could bring angular_components into the angular mono-repo. It would make it easier to keep them in sync and we could create some build tests from the example app.

The biggest con I see right now is that the development of the widgets is not centralized and changes come from lots of different product teams. Right now I do a lot of manual sanitizing of the code and commit descriptions to clean out internal product names and details. I wouldn't want to complicate the syncing of angular repos with that. When we work out a system to keep this clean, I'd revisit this idea.

[...]

Issue #602, opened 2017-08-29
API docs: OnInit has inlined code and no longer links to Lifecycle docs


chalin:
@matan: The APIs to some of the lifecycle hook interfaces were updated in this commit https://github.com/dart-lang/angular/commit/0f1abe5b9e2abfd537956271eead6ee6786c1b3c#diff-d9e2200e3226057f249e806a7e95d1adL76.

Consider OnInit as an example. Under [3.x, the OnInit documentation](https://webdev.dartlang.org/api/angular2/angular2/OnInit-class) has an "Example" section with

- A(n auto-updatable) code excerpt of (CI tested) code from the AngularDart doc examples
- A link to the AngularDart docs "Lifecycle Hooks" page.

In the [4.x OnInit API page](https://webdev-dartlang-org-dev.firebaseapp.com/api/angular/angular/OnInit-class), this has been replaced by inlined code block. Admittedly, the old example wasn't as good as it could be, but better ones can be found.

**My question is**: do we still want to stick to the old format of API docs along the lines of (the current 3.x docs):

> ## Examples
>
> Try the [live example](#) from the [Some docs page](#):
>
> \<?code-excerpt "docs/lifecycle-hooks/lib/spy_directive.dart (spy-directive)"?>
> [auto-updated codeblock]

cc @kwalrath

kwalrath:
Oops, this should be @matanlurey, not that other Matan...

chalin:
Oops indeed. Thanks for noticing!

PR #1395, opened 2018-06-11
chore: base setup for use of API doc code-excerpt updating tools

chalin:
This PR illustrates how to use two Dart tools to update code excerpts in API docs. The hope is that these tools can be used both internally and externally.

TL;DR

---

The two tools used to refresh code excerpts in the site-webdev and site-www docs are now written in Dart:

- https://github.com/chalin/code_excerpter is written as a builder. It scans source files for `#docregion` markers and creates `*.excerpt.yaml` excerpt files from the named regions.
- https://github.com/chalin/code_excerpt_updater can be run using `pub run`. Using the generated excerpt files, this tool scans the docs, refreshing any outdates excerpts as needed.

This PR illustrates the base setup for use of the tools. The tools are simply listed as `dev_dependencies` in the top-level `pubspec.yaml`.

This is how you can generate the code-excerpts from sources (both the local `examples/**` and the site-webdev example apps):

```console
pub upgrade 
pub run build_runner build --delete-conflicting-outputs --config excerpt --output tmp/excerpt
```

I've had to configure the builder to run over all packages, so the first build takes 22s (because some of the site-webdev example apps use ACX), but its 2s after that.

To update code excerpts in API docs, you can run commands like the following for `angular_forms`:

```console
pub run code_excerpt_updater \
  --no-escape-ng-interpolation \
  --fail-on-refresh \
  --write-in-place \
  --src-dir-path tmp/excerpt \
  --fragment-dir-path tmp/excerpt \
  --yaml angular_forms
```

You can update entire packages or individual files/folders:

```console
pub run code_excerpt_updater --no-escape-ng-interpolation --fail-on-refresh --write-in-place --src-dir-path tmp/excerpt --fragment-dir-path tmp/excerpt --yaml angular/lib/src/common/directives/ng_if.dart
```

I've run both of these commands and included the resulting changes in this PR.

Hopefully these two tools could be used both internally and externally, potentially contributing to solutions for the following issues:

- Validate that code snippets in docs are at least "analyzer clean" #996 
- Test & update/remove the examples in the API docs #1267
- Should docs and examples for webdev.dartlang live here? #942
- API docs: OnInit has inlined code and no longer links to Lifecycle docs #602
- Update documentation for Query.read to include an example #249

cc @alorenzen @matanlurey @kwalrath @kevmoo

### DIFF
--- a/angular/lib/src/common/directives/ng_if.dart
+++ b/angular/lib/src/common/directives/ng_if.dart
@@ -10,17 +10,33 @@ import 'package:angular/src/core/linker/app_view_utils.dart';
 ///
 /// ### Examples
 ///
-/// <?code-excerpt "docs/template-syntax/lib/app_component.html (NgIf-1)"?>
+/// <?code-excerpt "examples/hacker_news_pwa/lib/src/item_detail_component.html"?>
 /// ```html
-/// <hero-detail *ngIf="isActive"></hero-detail>
+/// <!-- Loading view -->
+/// <div *ngIf="item == null" class="notice">Loading...</div>
+///
+/// <!-- Loaded view -->
+/// <template [ngIf]="item != null">
+///   <item [item]="item"></item>
+///   <div *ngIf="item['comments'].isEmpty">There are no comments yet.</div>
+///   <comment
+///       *ngFor="let comment of item['comments']"
+///       [comment]="comment">
+///   </comment>
+/// </template>
 /// ```
 ///
-/// <?code-excerpt "docs/structural-directives/lib/app_component.html (asterisk)"?>
+/// <?code-excerpt "packages/template_syntax/app_component.html (NgIf-1)"?>
+/// ```html
+/// <my-hero *ngIf="isActive"></my-hero>
+/// ```
+///
+/// <?code-excerpt "packages/structural_directives/app_component.html (asterisk)"?>
 /// ```html
 /// <div *ngIf="hero != null" >{{hero.name}}</div>
 /// ```
 ///
-/// <?code-excerpt "docs/structural-directives/lib/app_component.html (ngif-template)"?>
+/// <?code-excerpt "packages/structural_directives/app_component.html (ngif-template)"?>
 /// ```html
 /// <template [ngIf]="hero != null">
 ///   <div>{{hero.name}}</div>
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,5 +5,24 @@
 name: repository
 
 dev_dependencies:
+  build_runner: any
+  code_excerpter:
+    git: https://github.com/chalin/code_excerpter.git
+  code_excerpt_updater:
+    git: https://github.com/chalin/code_excerpt_updater.git
   dev:
     path: dev
+
+  # Source of some of the code examples
+  structural_directives:
+    git:
+      url: https://github.com/dart-lang/site-webdev.git
+      path: examples/ng/doc/structural-directives
+  template_syntax:
+    git:
+      url: https://github.com/dart-lang/site-webdev.git
+      path: examples/ng/doc/template-syntax
+  pipes:
+    git:
+      url: https://github.com/dart-lang/site-webdev.git
+      path: examples/ng/api/common/pipes
--- a/angular_forms/lib/src/directives/ng_model.dart
+++ b/angular_forms/lib/src/directives/ng_model.dart
@@ -30,14 +30,14 @@ import 'shared.dart' show setUpControl;
 ///
 /// ### Examples
 ///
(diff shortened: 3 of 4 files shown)


angular/lib/src/common/directives/ng_if.dart
```diff
@@ -10,17 +10,33 @@ import 'package:angular/src/core/linker/app_view_utils.dart';
 ///
 /// ### Examples
 ///
-/// <?code-excerpt "docs/template-syntax/lib/app_component.html (NgIf-1)"?>
+/// <?code-excerpt "examples/hacker_news_pwa/lib/src/item_detail_component.html"?>
```
chalin: This illustrates use of code from `examples/hacker_news_pwa`.

pubspec.yaml
```diff
@@ -5,5 +5,24 @@
...
+  build_runner: any
+  code_excerpter:
+    git: https://github.com/chalin/code_excerpter.git
+  code_excerpt_updater:
+    git: https://github.com/chalin/code_excerpt_updater.git
   dev:
     path: dev
+
+  # Source of some of the code examples
+  structural_directives:
```
chalin: List site-webdev external apps here.

alorenzen:
@chalin thanks for sharing these. We have a docs fixit schedule for next week, where we're hoping to spend some time collating and improving our internal docs. After that, we should have a much better understanding of if / where we would want to use these tools.

alorenzen:
Thinking through how this would work in our internal/external workflow, here's what I've got: we basically treat examples like we do our golden files.

We can use `code_excerpter` as a test to tell us when the examples need to be updated. Since it is a builder, we can use the exact same approach as we did for the goldens.

We would then use `code_excerpt_updater` as the update script. We'd need a separate internal / external version like we have now for the goldens, but that shouldn't be too hard to handle.

We wouldn't use the "live snippet" feature of our internal markdown viewer, but this would still allow us to have our examples as actual examples that can be built / analyzed, and then embed them in the docs for both internal and external users.

@matanlurey does this sound reasonable to you?

I think an interesting question at this point is if we want to fork the tools and have them live in the angular repo, or if we want to make them official tools under dart-lang and just have wrappers in angular.
@matanlurey @kevmoo @kwalrath any thoughts on this?

kevmoo:
> I think an interesting question at this point is if we want to fork the tools and have them live in the angular repo, or if we want to make them official tools under dart-lang and just have wrappers in angular.

Could we start w/ these tools being in the angular repo and then graduate them to their own package/repo if we find other interested users?

chalin:
> Could we start w/ these tools being in the angular repo and then graduate them to their own package/repo if we find other interested users?

FYI, the tools are currently used by [dart-lang/site-www](https://github.com/dart-lang/site-www) as well as [dart-lang/site-webdev](https://github.com/dart-lang/site-webdev). @filiph is considering there use elsewhere too.

kevmoo:
Then we've passed my threshold for creating a repo: used by 2+ projects, has clear owner, plan to keep using to the horizon

matanlurey:
All of my thoughts below. We can schedule a meeting or start a doc to review further:

@alorenzen:

> Thinking through how this would work in our internal/external workflow, here's what I've got: we basically treat examples like we do our golden files.

I like what I'm hearing, and I think it's reasonable (your ideas, that is).

@kevmoo:

> Could we start w/ these tools being in the angular repo

@chalin:

> FYI, the tools are currently used by dart-lang/site-www as well as dart-lang/site-webdev. 

Unfortunately I don't feel comfortable using tools that we don't have direct control over. We've been burned and slowed down too many times in the past. If they lived in the AngularDart repo, _or_ we have direct control over what they do, at least until we have tests and integration in place, I'd be OK with making them more open/all over the place.

The reality right now is we need a way to be directly involved in keeping our examples and docs up to date, and not have to go through external folks or tools. However we can accomplish that is fine with me (and @alorenzen seems to have the right ideas, so I'll defer to him).

alorenzen:
@chalin 
> > Could we start w/ these tools being in the angular repo and then graduate them to their own package/repo if we find other interested users?

> FYI, the tools are currently used by dart-lang/site-www as well as dart-lang/site-webdev. @filiph is considering there use elsewhere too.

If we're going to use them internally, then I don't think the tools should live under https://github.com/chalin/

@kevmoo 
> Then we've passed my threshold for creating a repo: used by 2+ projects, _**has clear owner**_, plan to keep using to the horizon

I don't want individual owners. I want a team responsible for it. A team with OKRs that we can influence.

chalin:
> If we're going to use them internally, then I don't think the tools should live under https://github.com/chalin/

Totally agree; `dart-lang` seems like a good choice.

matanlurey:
I am going to close this because I think this PR will be stalled a bit until we have time for it.

(And we are still working out the details offline)

PR #1402, opened 2018-06-13
test(i18n): test translation of template messages

leonsenft:
Internalization (i18n) in templates is still a work in progress and will remain
disabled until it's ready for release.


leonsenft:
Not going forward with this approach. If we do want to support this we should use the tooling provided by **package:intl** with perhaps a hand written ARB file.

PR #1405, opened 2018-06-13
fix(Compiler): replace `retype()` with `cast()`

leonsenft:
The `retype()` method was removed in Dart 2.0.0-dev.62.0.


alorenzen:
The language team said they would be doing this work for us. Guess that never happened?

leonsenft:
CI is blocked by https://github.com/dart-lang/path/pull/42. I'm going to merge this internally anyways, since it should be backwards compatible with our Dart version constraints.

PR #1408, opened 2018-06-14
Run dartfmt fix over example apps to remove optional new/const.

alorenzen:


alorenzen:
Not sure if we want to actually merge this before dart2 is default in SDK?

PR #1409, opened 2018-06-15
Run dartfmt fix over entire angular repo to remove optional new/const.

alorenzen:
Just for comparison...


matanlurey:
Closing due to how stale this is (and it is easily auto-generated).

Issue #1038, opened 2018-03-07
We need a v4->v5 migration guide for external users


matanlurey:

eredo:
Currently working on something similar for us internally, but from v3.1 to v5, maybe I can contribute here. Where will it be stored?

kwalrath:
We have some info in https://webdev-dartlang-org-dev.firebaseapp.com/dart-2, but we'd be happy to point to anywhere we more info.

eredo:
@matanlurey  is this somewhat near to what you had in mind: https://github.com/eredo/angular/blob/doc-migration-v4-v5/doc/migration/v5.md ?

matanlurey:
@eredo Sorry it took so long to follow-up here. Why don't you send a PR for what you have so far?

eredo:
@matanlurey will do later today.

matanlurey:
Ping @eredo :)

matanlurey:
I'm happy with the doc and CHANGELOG we have today. It could always be improved, of course.

PR #1434, opened 2018-06-26
WIP: doc migration v4 v5

eredo:
This contains the first draft of the migration guide for v4 to v5 (#1038). Please let me know whether this is heading towards the right direction and I'll continue to add further details ASAP.


eredo:
I signed it!

doc/migration/v5.md
```diff
@@ -0,0 +1,184 @@
...
+dependencies:
+  # ...
+
+dev_dependencies:
+  # ...
+```
+
+Move your dart2js configuration out of your `pubspec.yaml` into a `build.release.yaml` config. 
+*This step is optional and the configuration may contain different configurations.*
+
```
eredo: I will reference to webdev as it already ships with a proper dart2js configuration.

doc/migration/v5.md
```diff
@@ -0,0 +1,184 @@
...
+```yaml
+transformers:
+- $dart2js:
+    checked: false
+    minify: true
+    commandLineOptions: [--show-package-warnings]
+```
+
+**New code(`build.release.yaml`):**
```
natebosch: I would drop references to `build.release.yaml` and suggest using `pub run build_runner --release` (or `-r`) instead of `--config release`. We now have built in support for using `dart2js` in "release" mode. Our intention is for most users to not need a `build.yaml` at all.

doc/migration/v5.md
jakemac53: Note that build_runner now also has a `--release` flag, and supports `release_options` and `debug_options` fields, as well as defaults that builders can expose for those.

Basically what that means is that `--release` will generally do the right thing and is probably what we should recommend going forward :). All of our builders already use that to for instance turn on dart2js, remove original Dart files, and remove source maps.

doc/migration/v5.md
```diff
@@ -0,0 +1,184 @@
...
+])
+final InjectorFactory injectorFactory = self.injectorFactory$Injector;
+
+void main() {
+  runApp(ng.AppComponentNgFactory, createInjector: injectorFactory);
+}
+```
+
+For further information about best practices for the dependency injector see 
+[effective DI](https://github.com/dart-lang/angular/blob/master/doc/effective/di.md#injectors).
```
jakemac53: Note that if using `runApp`, then you can also add this to your `build.yaml` and it will help ddc load times for things like angular_components a lot:

```yaml
global_options:
  angular|angular:
    options:
      no-emit-component-factories: True
      no-emit-injectable-factories: True
```

matanlurey:
Cursory LGTM. I'd be happy to land in the WIP state this is in after the above comments are addressed.

doc/migration/v5.md
```diff
@@ -0,0 +1,184 @@
...
+### Adjust analysis_options.yaml
+
+There are several cases where `.template.dart` files are included (ex. `main.dart` and async routes). Therefore adding
+the rule `uri_has_not_been_generated` to the `analysis_options.yaml` will prevent the analyzer from reporting 
+unnecessary issues.
+
+```yaml
+linter:
+  rules:
+    - uri_has_not_been_generated
```
hterkelsen: This code sample makes it look like we suggest turning on the `uri_has_not_been_generated` lint

doc/migration/v5.md
```diff
@@ -0,0 +1,184 @@
+# Angular Migration Guide v4 to v5
+
+## Major changes
+
+- Build system switched from pub transformers to build_runner/webdev. 
+  [Dart 2 Migration Guide for Web Apps](https://webdev-dartlang-org-dev.firebaseapp.com/dart-2)
+- Introduced Router 2.0. [Router Migration Guide](https://github.com/eredo/angular/blob/master/doc/router/migration.md)
```
leonsenft: Could you change this URL to point to the upstream Angular repository instead of your fork? I noticed your fork is out-of-date because this doc is missing Dart syntax highlighting in all of the examples.

doc/migration/v5.md
matanlurey: Yeah. This is not a lint, it's built into the analyzer. We should make sure this is clear.

doc/migration/v5.md
eredo: Great! I didn't know this, will check it out ASAP.

eredo: Amazing, saved ~50% load time! Adding it to the guide.

eredo:
Currently adding the other sections. Should be done within the next ~20 minutes.

eredo:
Okay this should be it for now. Most of the sections are referencing to existing documentation within the `doc` directory.

doc/migration/v5.md
jakemac53: Nice!

PR #1437, opened 2018-06-26
Support Event tearoffs in templates.

### LINKED ISSUE #506: Template syntax should support tear-offs for event bindings
Right now, we have to supply a method call for template bindings, using the magical `$event` variable.
`<material-button (trigger)="onClick($event)"></material-button>`

In order to be more "dart-y", and also to remove the need to remember the magic `$event` variable, we could instead support a tear-off
`<material-button (trigger)="onClick"></material-button>`

alorenzen:
Fixes #506.


_goldens/test/_files/directives/directive_wrapper.template.golden
```diff
@@ -67,10 +67,6 @@ class ViewTestFooComponent0 extends AppView<import1.TestFooComponent> {
     }
     _ChildDirective_0_5.detectHostChanges(this, _el_0);
   }
-
-  void _handle_trigger_0_0($event) {
-    ctx.onTrigger;
```
kevmoo: This just seems...weird. Did this ever do anything? Nothing caller or returned?

leonsenft: No, I believe this golden was just put in place so we could see the change.

PR #1443, opened 2018-06-29
don't build the debug entrypoints for tests since there are so many

jakemac53:
Related to https://github.com/dart-lang/build/issues/1620, this gives far fewer entrypoints to compile - but it does mean that you won't be able to debug tests in the browser using the original test file (you would have to go through --pause-after-load from the test package).

Feel free to close if you don't feel this is desirable.


matanlurey:
Could we add another `build.yaml` for this purpose, i.e. `build.debug.yaml`?

If so, SGTM.

jakemac53:
Sure I can push a `build.debug.yaml` to this branch which restores the old behavior. Note that it will end up invalidating builds if you switch back and forth though.

matanlurey:
Yeah that is fine. We'd use that only if we are debugging many tests in DDC.

PR #1458, opened 2018-07-07
chore(Misc): Cleanup analysis_options, naming.

matanlurey:
... remove `strong-mode: ...` which is deprecated, and `ng_generated` -> `ng`.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,5 +1,4 @@
 analyzer:
-  strong-mode: true
   errors:
     unused_element: error
     unused_import: error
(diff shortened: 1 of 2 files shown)


analysis_options.yaml
```diff
@@ -12,8 +11,8 @@ analyzer:
     # The noise of deprecations isn't worth the effort.
     deprecated_member_use: ignore
   exclude:
-    - .dart_tool/**
-    - build/**
+    - "*/.dart_tool/**"
+    - "*/build/**"
```
leonsenft: Did this not work in a specific scenario? Could you double check this still works in both IntelliJ and VS Code? See the last change I made here for context: https://github.com/dart-lang/angular/commit/d1fc081ab540e5d441a107874d5efe29657c17f2.

Issue #1454, opened 2018-07-06
Support named arguments, or generate a better error message


leonsenft:
Named method arguments aren't supported in templates. As reported by an external user, this can lead to some very misleading template parsing errors.

For example, the following snippet

```html
<material-input (ngModelChange)="item.copyWith(name: $event)" ...>
```

produces an error like

```sh
[SEVERE] angular on lib/src/item_component.dart:
Parser Error: Missing expected ) at column 22 in [item.copyWith(name: $event)] in 366
```

This error message gives no indication that the named argument is to blame. We should either consider supporting named arguments since they're an integral part of the language, or have an explicit error message.

While the latter sounds easier, simply identifying named parameters for error reporting would probably require most of the work necessary to support them altogether.

matanlurey:
The current parser is (still) based on the intersection between TypeScript and Dart, unfortunately, and I think lack of named arguments is directly related to that (we have _some_ differences, like `?.` operator). 

I don't know how trivial it would be to support named arguments here. If we do end up supporting them, we should probably also make sure it is supported elsewhere (dependency injection, for example, where it is not).

leonsenft:
@matanlurey Right I forgot we're not using `package:angular_ast` for expression parsing.

matanlurey:
And we deleted all the code in `angular_ast` related to expression parsing :(

leonsenft:
Yeah I remember now. We're going to be forced to implement something new anyway eventually. Perhaps we can address this issue then.

matanlurey:
Added to the >5.0 milestone. We should try and track this (and fix) before 6.0.

matanlurey:
So it looks like I could probably implement this...

I'll send a PoC out for review. We don't need to get it in though. The code isn't great, but it isn't worse than the existing parsing and output AST code (it basically just adds a few extra parameters to calls, and also will need a hack to solve for ambiguity around pipe syntax - which also uses `:` for arguments).

Issue #448, opened 2017-06-09
Support named arguments in dependency injection


matanlurey:
From @jakemac53:

Given a class like:

```dart
@Injectable()
class Foo {
  Foo({String bar});
}
```

The template compiler will end up generating a function like this in the `*.template.dart`:

```dart
(bar) => new Foo(bar);
```

It is a bit unclear what it should actually do in this case, possibly it should just be ignoring all optional args? Or maybe it should even throw, but that should happen during compilation.

matanlurey:
This is actually a breaking change, since existing constructors might have named arguments they expect never to be invoked via DI.

matanlurey:
We are closing as a duplicate of #553 - we decided adding that doing any sort of transitional work (errors, warnings, etc) should be handled as part of #553 and not by itself (i.e. here).

PR #1473, opened 2018-07-08
feat(Compiler): Adds support for named arguments.

matanlurey:
Closes https://github.com/dart-lang/angular/issues/1454.

Partial work towards https://github.com/dart-lang/angular/issues/448.

Not a priority to get in, we can discuss in person if this is the right approach.


matanlurey:
Merged.

Issue #1489, opened 2018-07-12
Add `castCallback1->N` to `package:angular/meta.dart`


matanlurey:
In AngularDart v5 with Dart2, the [lack of reified generics][1] definitely hurts:

```dart
typedef ItemRenderer<T> = String Function(T);

abstract class Dog {
  String get name;
}

class MyComponent {
  String itemRenderer(Dog dog) => dog.name;
}

class ItemListComponent<T> {
  ItemRenderer<T> itemRenderer;
}

void main() {
    var myComponent = new MyComponent();
    var childComponent = new ItemListComponent<dynamic>();

    //  TypeError: Type '(Dog) => String' is not a subtype of type '(dynamic) => String'
    childComponent.itemRenderer = myComponent.itemRenderer;
}
```

We're going to work on [#68][1], but until then, users need an _ergonomic_ way to _untype_ these types of functions/closures, but _retain_ the some benefits of static typing (IDE auto-completion, go-to definition, avoid dynamic calls).

My proposal is to add a `package:angular/meta.dart` ([we were already planning this for `@visibleForTemplate`][2]), and include a few helper functions to cast callbacks into forms that are suitable for AngularDart components, with the goal of removing/fixing these cases when we fix [#68][1]:

```dart
T Function(dynamic) castCallback1<T, A>(T Function(A) callback) {
  return (dynamic element) => callback(element as A);  
}
```

You can see a full proposal, and example, on DartPad:
https://dartpad.dartlang.org/82e79cbcd0c781b9cda557754fbccbf8

(I am proposing this for v5 final, as a migration aid from v4)

Another advantage is we can leverage `unsafeCast` (internal function) to reduce the amount of runtime casts that occur in production code, retaining performance from v4 (and Dart 1) as much as possible.

[1]:  https://github.com/dart-lang/angular/issues/68
[2]: https://github.com/dart-lang/angular/blob/09ed555b1a75141406cb96e54f9807e704bed875/angular/lib/src/meta.dart

alorenzen:
Not sure why it needs to live in angular, put understand the reasoning for having it somewhere.

Maybe we could just stick it in quiver?

matanlurey:
We don't have a dependency on quiver (intentionally, it breaks too often). The reason for Angular is to track the usages of it, and once we have proper generic support we could `@deprecate` it with references on how to use template generics.

PR #1497, opened 2018-07-17
Fix typo in issue number, and add issue number

chalin:
- Fix typo in issue number: `#1849` -> #1489
- Add a reference to: Template syntax should support tear-offs for event bindings #506

cc @kwalrath

### DIFF
--- a/angular/CHANGELOG.md
+++ b/angular/CHANGELOG.md
@@ -37,7 +37,7 @@
     unresolved.
 
 *   Support inherited method tear-offs in event listeners. Previously, we only
-    considered methods in the Component class itself.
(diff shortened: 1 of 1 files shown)


angular/CHANGELOG.md
```diff
@@ -57,12 +57,13 @@
     produced accidentally created two queries instead of one.
     [Closes #1455][#1455].
```
chalin: @alorenzen: in general when referring to an issue number, how about **simply writing the issue number in parentheses** at the end of the bullet rather than `Closes #1234`? It would be more concise and amount to less typing.

PR #1515, opened 2018-07-22
GitHub Sync

matanlurey:
This is an automatically generated pull request.


kevmoo:
screen shot 2018-07-27 at 7 41 31 pm

Did source_gen get configured correctly? I'm seeing a bunch of `.g.dart` files after trying out the hacker_news build.

@natebosch ?

leonsenft:
@kevmoo [pkg:pageloader](https://pub.dartlang.org/packages/pageloader/versions/3.0.0-beta) generates the page objects in the source tree. It looks like somehow the .gitignore change wasn't committed... strange. I wonder if our Copybara script ignored it.

.gitignore
```diff
@@ -1,10 +1,9 @@
...
-.dart_tool/
 pubspec.lock
 
 # Files generated by dart tools.
-.dart_tool
-/doc/api
-**/doc/api
+**/.dart_tool/
+**/doc/api/
+*.g.dart
```
kevmoo: Maybe we should do `*_po.go.dart` instead? I could imagine using generated parts in other areas of this repo in the future...

leonsenft: We already do for offline_urls.g.dart.

This doesn't prevent you from tracking or committing a `.g.dart` file, it just want show as unstaged _until_ you add it to be tracked.

PR #1517, opened 2018-07-23
use build_vm_compilers for incremental kernel file generation

jakemac53:
Also changed to allow the latest mockito so that the latest package:test can be picked up.


jakemac53:
Looks like for a fresh build this adds ~2 mins to build time, but it takes ~2 mins off the test run time, so essentially a wash but then you get the benefit of modular compilation when the build is not invalidated.

PR #1544, opened 2018-08-06
Update pubspecs for dart2.0 and angular 5 stable

alorenzen:
Note, I still need to update the CHANGELOGs. But running this through travis while I work on those.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ language: dart
 
 dart:
   - stable
-  - dev
 
 # Gives more resources on Travis (8GB Ram, 2CPUs).
 # Do not remove without verifying w/ Travis.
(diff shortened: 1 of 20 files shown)


kevmoo:
Leave the SDK lower-bound constraints as-is. There may be folks unable to roll to stable due to dependency issues.

.travis.yml
```diff
@@ -48,6 +48,7 @@ jobs:
   # See https://github.com/dart-lang/angular/issues/1114.
   allow_failures:
     - env: PKG="examples/github_issues"
+    - env: PKG="examples/hacker_news_pwa"
```
kevmoo: This doesn't seem to affect anything...

alorenzen: Yes it does. It allows travis to continue to follow-up stages even if this one job had failures.

alorenzen: `Your build matrix was set to allow the failure of job 5044.14 so we continued this build to the next stage.`

kevmoo:
I was used to something showing up on the display to show 'allow_failures'
– just not w/ build steps?

Issue #1421, opened 2018-06-18
AppView -> {Component|Host|Embedded}AppView


matanlurey:
Placeholder cleanup issue. Basically, the reason for this documentation being necessary:

* https://github.com/dart-lang/angular/blob/master/doc/generated_code.md#host-view
* https://github.com/dart-lang/angular/blob/master/doc/generated_code.md#component-view
* https://github.com/dart-lang/angular/blob/master/doc/generated_code.md#embedded-views

PR #1586, opened 2018-08-17
WIP: Imagine a world with split AppView classes...

matanlurey:
Work towards https://github.com/dart-lang/angular/issues/1421.

### DIFF
--- a/angular/lib/src/runtime/view.dart
+++ b/angular/lib/src/runtime/view.dart
@@ -21,7 +21,9 @@ abstract class View<T> {
 
   /// Destroys this view and disposes of any resources or child views.
   @mustCallSuper
-  void destroy() {}
+  void destroy() {
+
+  }
 }
 
 /// Base generated view for views that have a parent [ComponentView].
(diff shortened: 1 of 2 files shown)


angular/lib/src/runtime/view.dart
```diff
@@ -0,0 +1,176 @@
...
+/// ... would generate:
+///
+/// ```dart
+/// class _EmbeddedView_0_NgIf$A extends EmbeddedView<A> {}
+/// ```
+abstract class EmbeddedView<T> extends ChildView<T> {
+  @protected
+  final ComponentView<T> componentView;
+
+  final Map<String, dynamic> _locals = Map.identity();
```
leonsenft: Could we possible type `_locals` with a type parameter?

```dart
abstract class EmbeddedView<T, L> extends ChildView<T> {
  final L locals;
  ...
}
```

Then invocations of `embeddedView.getLocal(String name)` could be replaced with `embeddedView.locals.name`.

It's not clear to me whether or how this works in every context we need to invoke locals. For example, anything that accesses a local from the template would probably be covered, however how would client code get the `L` type? Would they inject a `Template<L>` or `ViewContainer<L>`?

PR #1617, opened 2018-09-18
Update migrating-to-v5.md

drbcode:


leonsenft:
Thanks! We'll pull this in the next sync.

PR #1618, opened 2018-09-18
Update issue-labels.md

drbcode:


leonsenft:
I realize the note contradicts the fact that the file indeed is formatted to 80 characters, but there's really no reason to stop formatting it now that it already is formatted.

PR #1632, opened 2018-10-08
GitHub Sync

alorenzen:
This is an automatically generated pull request.


leonsenft:
Looks like we need to bump the minimum SDK version: https://travis-ci.org/dart-lang/angular/jobs/438840055 (Future is exported by dart:core in 2.1.0-dev...). I'll send out a fix.

alorenzen:
I don't think we want to require another dev SDK just yet. Can we instead keep importing `dart:async`?

PR #1713, opened 2019-02-01
Rename OnDeactivate's parameters

TastyPi:
"previous" and "current" are very confusing names for this method, especially because they mean different things to OnActivate's previous and current. Using "current" and "next" instead makes it clearer that the first argument is for the current Component's RouterState and "next" is for the next component.

This also matches the code in router_impl where _activeState and nextState are passed to onDeactivate.


TastyPi:
All the build failures have nothing to do with this change.

PR #1724, opened 2019-02-24
Funnel developers to the larger example repository

exo-pla-net:


matanlurey:
Thanks @exo-pla-net. Unfortunately I don't feel comfortable linking to other repositories/orgs that we do not personally maintain or review, otherwise folks will get the wrong idea or be frustrated when something is out of date.

PR #1844, opened 2019-07-25
allow to use parent route in RouteDefinition

sinegovsky-ivan:


angular_router/lib/src/route_definition.dart
```diff
@@ -177,7 +180,7 @@ abstract class RouteDefinition {
     if (isDevMode && paramValues == null) {
       throw ArgumentError.notNull('paramValues');
     }
-    var url = '/' + path;
+    var url = '/' + (_routePath != null ? _routePath.toUrl() : path);
```
leonsenft: There's an inconsistency in the logic here.

In the constructor, `path` is defined as `path ?? routePath?.path`, meaning the `path` takes priority there. Whereas `routePath` takes priority here.

This is a potential for bugs. I think we'd need to update the constructor to prefer `routePath.path` over `path` as well.

PR #1917, opened 2020-10-01
Rename dart_test.yaml to dartTest.yaml

meetmeofficial0602:
It's good to follow the camelCase naming convention


kevmoo:
For file names, we use `x_y` – so it works fine on OSes that don't handle case well.

Also: this file name is specific to `pkg:test`

